PR #2, opened 2018-03-28
Add weights to gridder/degridder and remove psf function

sjperkins:
Gridding visibilities and the PSF happen in separate functions. However, in Ben's original formulation, they were gridded within the same function. I've returned the code to this setup as

1. It's probably faster.
2. Less code duplication.

I originally split the two out too make dask code easier to read, but I don't think this is an overriding concern any longer.


africanus/gridding/simple/gridding.py
```diff
@@ -43,142 +54,96 @@ def _grid(vis, uvw, flags, ref_wave, nx, ny, convolution_filter):
...
-    #    (A) apply beam DDE gains
-    #        Sample the facet centre position
-    #        in the beam cube as it varies over time
-    #        to produce these gains.
-    #    (B) average
-    # 2. Foreach row in averaged block
+    # for deconvolution the PSF should be 2x size of the image (see
+    # Hogbom CLEAN for details), one grid for the sampling function:
+    # Reference the real component of the PSF, imaginary is zero
+    psf = np.zeros((2*ny,2*nx), dtype=np.complex64)
```
landmanbester: True, ideally you need a PSF that is twice the size of the image but we often want speed things up by using CLEAN boxes (known as Clark CLEAN). You might want to add an input parameter which specifies how large the PSF should be (as a fraction of the image size for example). Also, when running multiple major cycles, you do not want to recompute the PSF every time. So you should also add a parameter that specifies whether the PSF should be computed or not. Having said that, I don't see the weight column being passed in. Are we currently assuming that all visibilities are equally weighted?

africanus/gridding/simple/gridding.py
```diff
@@ -43,142 +54,96 @@ def _grid(vis, uvw, flags, ref_wave, nx, ny, convolution_filter):
...
-            scaled_u = uvw[r,1] / ref_wave[c]
+                # Iterate over u/x
+                for conv_u in filter_index:
+                    base_u = conv_u*cf.oversample
+                    u_tap = cf.filter_taps[base_u + frac_u]
+                    u_tap_psf = cf.filter_taps[base_u + frac_u_psf]
 
-            disc_u = int(round(scaled_u))
-            disc_v = int(round(scaled_v))
+                    conv_weight = v_tap * u_tap
```
landmanbester: Looks like we are assuming a WEIGHT column of all ones. We should incorporate a weighting scheme. I think just natural weighting will be fine, at least for what I am planning to use the gridder for. But that is a separate pull request...

sjperkins:
I've

1. Added weights
2. Removed PSF calculation from the gridding function
3. Tested
  - Created two PSF's with size (ny, nx) and (ny*2, nx*2) by gridding with natural weights and unity visibilities.
  - Ensured that the centre of the bigger PSF equals that of the smaller PSF.

PR #29, opened 2018-06-05
Added division by n to dft kernels

landmanbester:
Simple change, just added division by n to im_to_vis and vis_to_im kernels. Let's see if I broke anything in the process

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,7 +5,7 @@ History
 0.1.3 (2018-03-28)
 ------------------
 
-* Added division by :math:`n` to DFT (:pr:`29`)
+* Added division by :math:`n` to DFT. Fixed dask chunking issue. Updated test_vis_to_im_dask (:pr:`29`)
 * Implement RIME visibility predict (:pr:`24`, :pr:`25`)
 * Direct Fourier Transform (:pr:`19`)
 * Parallactic Angle computation (:pr:`15`)
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,6 +5,7 @@ History
 0.1.3 (2018-03-28)
 ------------------
 
+* Added division by :math:`n` to DFT (:pr:`29`)
 * Implement RIME visibility predict (:pr:`24`, :pr:`25`)
 * Direct Fourier Transform (:pr:`19`)
 * Parallactic Angle computation (:pr:`15`)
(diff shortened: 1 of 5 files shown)


sjperkins:
How about documenting the divide by n in the maths docstrings?

landmanbester:
Alright, as discussed with @sjperkins I added some docs for the DFT. @o-smirnov I'm curious to hear your opinion on what I wrote there, I think it might explain why we've been struggling to get resolve working in DDF...

sjperkins:
@landmanbester Could you also add the PR details to HISTORY.rst?

Should I hold off merging until @o-smirnov has commented?

landmanbester:
Done. Maybe hold off until I've tested it inside a deconvolution algorithm, will try to get to that today

landmanbester:
@sjperkins we need to hold off before merging this in. For some reason the dask version seems to only be using the result for a single. Checked this by computing the PSF and found that the max of the PSF is always equal to the chunk size. No clue why this is happening though. Any ideas?

sjperkins:
@landmanbester Are you referring to the PSF calculated by `grid`? If so, yes, it produces a dask array of a single chunk because the gridder needs random access to the entire image domain. I don't know if gridding over sub-domains of the image would be useful, but lets discuss in person.

For the moment, I see the output of `vis_to_im` has shape `(source, chan)`.  I *think* something like this will make the chunks compatible as it will produce a single chunk

```python
I = vis_to_im(...)
I = I.rechunk((npix, npix, nchan))
I = I.rechunk(I.shape)
```

landmanbester:
No I'm making the PSF with vis_to_im. I think this might have something to do with the += inside the the vis_to_im kernel. The image looks fine but its clearly made from the visibilities of a single chunk. The numpy version works as expected. The tests miss this because the dask test uses only a single visibility chunk

sjperkins:
Ah I see. You probably need something like [this](https://github.com/landmanbester/codex-africanus/blob/2ac725372bf980903b2f8fe0bec7ee89498252e4/africanus/gridding/simple/dask.py#L39-L52). Note the extra `'row'` dimension in the output, the `adjust_chunks={'row':1}` and the reduction over the `row` dimension in the return statement.

landmanbester:
Awesome, thanks @sjperkins, that seems to have done the trick. Now to see if we can actually use it for deconvolution...

sjperkins:
Cool, just remember to push the commit up!

sjperkins:
@caramel-koala The DFT has now been merged into master branch, so it should hopefully be a bit easier for you to run your applications now.

PR #30, opened 2018-06-19
Zernicke DDE dask implementation

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  remaining issues.

  ```
  $ autopep8 -r -i africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```


sjperkins:
@JoshVStaden The `zernike_dde` function has been updated to support multiple correlations by adding extra dimensions to `coeffs` and `noll_index`. They now have shape `(ant, chan, corr_1,..., corr_n, poly)`. I've added a test case `test_zernike_multiple_dims` to check that this kind of works.

Could you and @landmanbester take a quick look and give me an indication if I've done the right thing?

JoshVStaden:
@sjperkins Hi Simon, in terms of adding correlations into the code, I'm not sure if each correlation needs to be added together. From what I saw in Eidos, the output itself has the extra dimensions as well for the different correlations. When I tested the different correlations, I tested each one separately.

sjperkins:
@JoshVStaden There are separate correlation dimension(s) in the output now if you take a look at `test_zernike_multiple_dims`

PR #32, opened 2018-07-10
[WIP] Primal Dual

caramel-koala:
Don't actually pass this pull request, just so you can see the issue, it's in tests/test_opts from line 50.

The issue is that it only runs on a single core on my machine and also the warning:
UserWarning: Increasing number of chunks by factor of 400 (nparts / max_parts))

I assume this has to do with the chunking being misaligned, but a multiple of each chunk fits completely in the array it's assigned to or ¯\_(ツ)_/¯


sjperkins:
@caramel-koala  Could you please

```python
print(a.name, a.numblocks, a.chunks)
```

for each of the dask arrays passed into `vis_to_im`.

Could you also do the same for the dask array produced by `vis_to_im` (you should remove the compute()) call at the end.

sjperkins:
Additionally, a chunk size of 25 is unlikely to exercise your cores -- its far too small. Split the rows into chunks of 10,000 to 50,000 and see what happens...

caramel-koala:
The larger array sizes fixed it.

caramel-koala:
Got a new problem though, when I call the dft (non dask as I'm still testing the primaldual itself) functions from within the power method (in sub_opts), which are defined in test_opts as L and LT, I get a

Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)

When I remove the numba.jit from the dft this goes away. The weird thing is that when I call L or LT from anywhere else it works fine, just when I call it from pow_method does it give this error.

Any idea as to why, @sjperkins?

sjperkins:
/cc @landmanbester @caramel-koala 

For future reference, one can add `debug=True` to the jit decorator. This'll produce machine code debuggable with gdb, showing the line which segfaulted.

sjperkins:
It's probably worth putting some shape checks which throw exceptions in the jitted code for these kind of errors. It's not possible to format the exception string, but "'source' dimension does not match" is better than nothing.

caramel-koala:
Turns out it's because I was giving it a square image instead of a flat one, I've made the requested changes, will push them once the tests are done.

sjperkins:
Might as well keep the PR open for further queries during development

sjperkins:
@caramel-koala Probably just need to merge master branch in to fix things up.

landmanbester:
I think this project is dead, closing for now

Issue #34, opened 2018-07-24
Using real phase in im_to_vis 


landmanbester:
Well, this is embarrassing. Somehow I forgot to add the factor of 1.0j to the real phase in im_to_vis. The test missed this because it only checks if im_to_vis produces the correct thing for a source at the phase centre. I will add a more thorough test with the pull request

sjperkins:
Its a remnant of the sin/cos decomposition for purely imaginary numbers ;-)

PR #35, opened 2018-07-31
Fixed bug in im_to_vis

landmanbester:
Fixes https://github.com/ska-sa/codex-africanus/issues/34

I also removed the division by n in the dft since this unnecessarily complicates things (e.g. makes the PSF direction dependent in a trivial way).

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,7 +4,9 @@ History
 
 0.1.3 (2018-03-28)
 ------------------
-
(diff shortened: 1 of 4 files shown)


sjperkins:
Just fix the failing test case ;-)

landmanbester:
The operators are now also consistent to within 1e-12 so I lowered the tolerance for the adjointness test to 1e-11

PR #38, opened 2018-08-02
W-Stacking

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  remaining issues.

  ```
  $ autopep8 -r -i africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```


sjperkins:
TODO
--------

- [ ] Cleanup filter + taper interface. Should be some consistency here...

PR #42, opened 2018-08-27
Add the simple dask gridder example

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  remaining issues.

  ```
  $ autopep8 -r -i africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,7 @@ History
 
 0.1.3 (2018-03-28)
 ------------------
+* Add simple dask gridder example (:pr:`42`)
 * Implement Kaiser Bessel filter (:pr:`38`)
 * Implement W-stacking gridder/degridder (:pr:`38`)
 * Use 2D filters by default (:pr:`37`)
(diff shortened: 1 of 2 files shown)


sjperkins:
@bennahugo I had this script lying around, nice to include it alongside the non-parallel version. I've assigned you as a reviewer for your interests sake, don't worry about a thorough look.

You'd probably find the contrast between the dask and the numpy version interesting...

- [dask](https://github.com/ska-sa/codex-africanus/blob/cfa63a02ede8ce1bbefb88ffa60f56461285bdf3/examples/gridding/simple/dask_grid.py)
- [numpy](https://github.com/ska-sa/codex-africanus/blob/cfa63a02ede8ce1bbefb88ffa60f56461285bdf3/examples/gridding/simple/toygrid.py)

sjperkins:
> is there any difference between the delayed operator and the tensor operator when wrapping functions aside from specifying labeled axes

The delayed (dask?) operator wraps the tensor (numpy?) operator. However, in the numpy case, one grid is allocated and passed through for each chunk of visibilities (`grid=grid`). In the dask case, a separate grid is created for each chunk of visibilities, but these separate grids are summed in a parallel reduction.

PR #43, opened 2018-08-31
Fix gridding examples

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  remaining issues.

  ```
  $ autopep8 -r -i africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/examples/gridding/simple/toygrid.py
+++ b/examples/gridding/simple/toygrid.py
@@ -24,14 +24,14 @@ def create_parser():
     p.add_argument("ms")
     p.add_argument("-rc", "--row-chunks", default=10000, type=int)
     p.add_argument("-np", "--npix", default=1024, type=int)
-    p.add_argument("-sc", "--cell-size", default=6, type=float)
+    p.add_argument("-sc", "--cell-size", type=float)
     return p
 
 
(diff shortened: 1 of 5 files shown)


sjperkins:
Makes sure that the simple and w-stacking gridding examples have the same inputs.

- Same cell size
- Same convolution kernel sizes
- Also removed unncessary discretisation of W for the simple gridder. W isn't even used by the simple gridder.

examples/gridding/simple/toygrid.py
```diff
@@ -174,4 +188,4 @@ def create_parser():
 
         # Produce visibilities for this chunk of UVW coordinates
         vis = degrid(dirty, uvw, natural_weight, wavelength,
```
landmanbester: Glad to see weights in both the gridder and degridder but need to confirm this is doing what I hope it's doing. Weights are usually only applied during gridding but this breaks the self-adjointness of the operator. Better to apply the sqrt of the weights going both ways. Is this what's happening here?

PR #48, opened 2018-09-07
Fix kaiser bessel filter and taper

sjperkins:
- Fix Kaiser Bessel filter by multiplying through by the `sinc` function.
- Numerically compute the taper of the Kaiser Bessel.
- Add plot-filter and plot-taper command-line utilities.

- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  remaining issues.

  ```
  $ autopep8 -r -i africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,7 @@ History
 
 0.1.3 (2018-03-28)
 ------------------
+* Fix Kaiser Bessel filter and taper (:pr:`48`)
 * Stokes/Correlation conversion (:pr:`41`)
 * Fix gridding examples (:pr:`43`)
 * Add simple dask gridder example (:pr:`42`)
--- a/africanus/filters/conv_filters.py
+++ b/africanus/filters/conv_filters.py
@@ -73,8 +73,7 @@ def convolution_filter(half_support, oversampling_factor,
         Beta shape parameter for
         `Kaiser Bessel <kaiser-bessel-filter_>`_ filters.
     normalise : {True, False}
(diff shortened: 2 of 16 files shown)


sjperkins:
@landmanbester She's all yours

landmanbester:
Ok so I have been testing both the simple gridder and the degridder against the DFT (which I am pretty sure of because it agrees with Meqtrees). Here is the fractional difference between two dirty images obtained from random visibilities and uv-coverages with w set to zero

dirty_diff

For the degridder I get a max difference of 1e-6 for a point source at the centre and it gets worse as I move it away from the centre. The script I used to compare the two is [here](https://github.com/landmanbester/codex-africanus/blob/fix-kaiser-bessel-filter-and-taper/africanus/gridding/simple/tests/gridder_vs_dft.py). Any thoughts on what could be causing this @bennahugo @o-smirnov? Looks like it might be the taper but not sure how since I'm using the standard formula for the Kaiser Bessel taper

landmanbester:
This definitely has something to do with how the filter/taper is being applied. Here is the difference if I make both the oversampling and support even

dirty_diff_even

Seems to make it worse in the opposit direction....

sjperkins:
@landmanbester I can't offer much input on the discrepancy between the DFT and the gridder/degridder. However, I do want to say that the `test_forward_operator` and `test_backward_operator` tests are a good axiom for any future gridders in codex.

PR #50, opened 2018-10-29
Enhancements to visibilty prediction

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  remaining issues.

  ```
  $ autopep8 -r -i africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```


sjperkins:
@landmanbester OK, I've bashed the test suite into working order.

There were multiple things going wrong which I've fixed by:

1. `python setup.py test` in the docker files was causing very weird issues on python 2.7 (numpy umr_sum() failures in some of the earlier commits. I've resolved this by calling `py.test /code/africanus` instead.
2. We only do python-casacore (parallactic_angle tests) on python2. python-casacore is broken on python3 as https://github.com/casacore/python-casacore/pull/132 hasn't made it into a release yet.
3. numpy 1.15.3 was causing issues with astropy https://github.com/astropy/astropy/pull/7955.
4. The latest dask 0.20.0 was also causing issues https://github.com/dask/dask/issues/4038#issuecomment-434338006.

The test cases and the numba predict_vis function are probably the important things to take a look at.

africanus/rime/predict.py
```diff
@@ -167,20 +498,23 @@ def predict_vis(time_index, antenna1, antenna2,
...
-        shape :code:`(source,time,ant,chan,corr_1,corr_2)`
-    ant2_jones : :class:`numpy.ndarray`
-        Per-source Jones terms for the second antenna.
+    ant1_jones : :class:`numpy.ndarray`, optional
+        :math:`A_{ps}` per-source Jones terms for the first antenna.
         shape :code:`(source,time,ant,chan,corr_1,corr_2)`
-    row_jones : :class:`numpy.ndarray`
-        Per-source coherency matrix for the row (baseline)
+    bl_jones : :class:`numpy.ndarray`, optional
+        :math:`X_{pqs}` [er-source coherency matrix for the row (baseline).
```
landmanbester: Typo er-source -> per-source. Same typo also below

PR #79, opened 2019-02-25
Compare codex-africanus predict vs meqtrees

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```


sjperkins:
@paoloserra This PR started testing the `radec_to_lm` conversion vs astropy to ensure they agree. 

I extended it to add an end-to-end test comparing the codex-africanus predict against meqtrees. They also agree once one takes [sign conventions](https://github.com/ska-sa/codex-africanus/pull/79/commits/4d04bdc8d0e4d8d49ca3edec36ce3dfb813df55c) into account, but I've been told that this shouldn't matter for the purposes of imaging. The script does have a `--invert-uvw` command now to facilitate this.

I've also updated the predict script to automatically take the correlation types associated with each `DATA_DESC_ID` into account so one shouldn't need to manually specify linear/circular feeds any longer.

I'm going to merge this PR now. Let us know whether the predict is successful for your subtraction purposes.

paoloserra:
Thanks @sjperkins , this worked very well.

I've had to use the -iuvw option to place the model components at the correct Ra,Dec.

I've tested it both for a single source and for a collection of flat-spectrum, unpolarised sources. It did everything correctly.

I will now implement more complex model spectral shapes based on the WSclean output.

sjperkins:
> Thanks @sjperkins , this worked very well.

Cool, one thing you might try is setting the row_chunk and source_chunk sizes higher. This will probably increase your core usage at the expensive of increasing memory -- there's a tradeoff here.

> I've had to use the -iuvw option to place the model components at the correct Ra,Dec.

Hmmmm, I did not expect you would have to use this option to get the correctt results. I should probably do some more investigation here as I would like everything to "*just work*".

Let me know if Gaussian Shape parameters would be useful for you because I think they are easy to add.

paoloserra:
> Let me know if Gaussian Shape parameters would be useful for you because I think they are easy to add.

Yes they would be. In some fields we must use multi-scale clean. The model will then include Gaussian components.

PR #80, opened 2019-02-28
Full Time and Channel Averaging Implementation

sjperkins:
This updates the Time and Channel Averaging code to average baselines in intervals determined by the TIME and INTERVAL. These columns contain the *nominal* data while most other columns contain *effective* data. Quoting from the [MS v2.0](https://casa.nrao.edu/Memos/229.html) specification:

TIME

    Mid-point (not centroid) of data interval.

INTERVAL

    Data sampling interval. This is the nominal data interval and does
    not include the effects of bad data or partial integration.

TIME_CENTROID

    The TIME column in MAIN is defined as the mid-point of the nominal
    sampling interval, as specified in the column, INTERVAL. The EXPOSURE
    column, in contrast to INTERVAL, defines the effective integration time,
    including the effects of missing data.`

EXPOSURE

    Effective data interval, including bad data and partial averaging.

To support this specification, the averager handles the distinction between these column pairs

1. TIME/INTERVAL averaging bins include both flagged and unflagged data.
2. TIME_CENTROID/EXPOSURE averaging bins never include flagged data, *except* if the entire bin is flagged in which case the bin is mark as flagged.

Visibilities are averaged by a mean of circular quantities.

- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```


sjperkins:
@mulan-94 @SpheMakh its in, go forth and break it.

Docs: https://codex-africanus.readthedocs.io/en/latest/averaging-api.html

PR #99, opened 2019-04-04
Shapelets

JoshVStaden:
- [ ] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```


JoshVStaden:
@sjperkins Hi Simon, I see that adding in a ndmin=1 option under loadtxt seems to force the output to have at least 1 dimension. The problem is that now, for some reason, the script has decided that I don't have cupy, and it keeps throwing an error saying that I don't have it. I tried pip installing it on the Jake server, and it said that there is no CUDA on the system. Not too sure how to get around this.

sjperkins:
@joshvstaden ok, well your dask wrapper is wrapping the cupy outline that you've written in africanus.model.shapes.shapelets.py

Perhaps you're intending to wrap the implementation in africanus.rime?

At some point, it would be good if this was moved into africanus.model.shape for consistencies sake.

JoshVStaden:
Maybe I should just take the cupy implementation out for now, just for simplicity's sake. Will still probably need to install cupy at some stage, though.

JoshVStaden:
I am wrapping the implementation in africanus.model.shape. I forget, what is the difference between the one in africanus.rime and in africanus.model.shape?

JoshVStaden:
Oh, looking at the file now, I see what's wrong. I seem to have copied the wrong file across.

sjperkins:
@JoshVStaden well a cuda implementation is worthwhile, it's just going to fail if cupy isn't installed. That's what the requires_optional decorator is for (read it's docs!).

However, to be consistent with the rest of africanus, the dask layer should wrap the numba implementation. I suspect you actually want to wrap the implementation in africanus.rime.shapelets.py. 

I think there's also confusion here because there was talk of shapelets for DDEs, and shapelets which describe the shape of a radio source. These are two separate things and I think we're mostly interested in the second for your project.

@landmanbester, can you confirm the above?

JoshVStaden:
I put the correct shapelet file in the correct directory, and then changed the rime shapelets to shapelet_dde.py, just to try to avoid confusion. There is now shapelet, which computes the model image shapelets, and shapelet_dde which computes the dde shapelets.

JoshVStaden:
@sjperkins @landmanbester Hi. I checked the visibilities on the MODEL_DATA column in the measurement set, and they were all real-valued. Also, Landman, just to confirm, by zero-coefficient shapelet, do you mean just the shapelets that I am generating? As in just a single coefficient of 1, and just beta values of 1?

landmanbester:
Sorry no, I mean the zeroth order coefficient i.e. the first one in the expansion. Since the zeroth order Hermite polynomial is just one, the corresponding shapelet is a Gaussian, at least up to a normalisation factor. Now, since the Gaussian is symmetric, the corresponding visibilities will all be real. The fact that MODEL_DATA is all real is encouraging. Did you image it yet?

JoshVStaden:
Yea, the image isn't showing anything.

landmanbester:
That's what I also saw. Imaging the CORRECTED_DATA (or DATA I am not sure now) column I get something but not what is expected. Which column are you writing to during the predict?

JoshVStaden:
I'm writing to the MODEL_DATA column

JoshVStaden:
Hi, I attempted to generate the graph of baseline vs visibility amplitude. It seems that the real part of the visibilities either take on the value of 0.5641896 or 0, and there's no in-between. I generated a scatterplot of the baseline length vs the visibility and, as expected, it didn't taper off as we wanted it to. At 0 baseline length, it took on that value of 0.5641896, and, for the remainder, it stayed at 0. I just wanted to verify that the code I used to generate the graph is correct:

`
        plt.figure()
        plt.scatter((np.sqrt(uvw[:, 0] **2 + uvw[:, 1] **2)), np.real(np_model[:, 0, 0]))
        plt.savefig('vis_vs_baseline.png')
        plt.close()
`

I have also uploaded the file to GitHub. I was wondering if there was any significance in the visibilities ranging between 0 and some strange value.

o-smirnov:
Yep, the plotting snippet looks correct.

> I have also uploaded the file to GitHub. I was wondering if there was any significance in the visibilities ranging between 0 and some strange value.

Clearly the code is computing the wrong thing then... no point trying anything more complicated until we get this right. You should see a smooth tapering down with increasing baseline. Maybe somewhere the scaling of the u, v coordinates is so wrong that it tapers down to 0 immediately?

JoshVStaden:
Yes, must be. I noticed my code seems to be used to using a full uv coverage (from the coordinates generated by my own test script), and the measurement set gives a semicircle, and with a completely different range of values. I'll take a look into how these coordinates affect my code.

JoshVStaden:
Okay, I think I see the issue now. It seems the uv values get too large for my code to handle. According to python, if I input np.exp(-92 ** 2), for example, it just outputs 0. Is there some sort of numpy function I can use to force my code to use more bits to store the number?

sjperkins:
>  and the measurement set gives a semicircle

WSRT (Westerbork Synthesis Radio Telescope) has an East-West antenna configuration, so this is to be expected.

> Okay, I think I see the issue now. It seems the uv values get too large for my code to handle. According to python, if I input np.exp(-92 ** 2), for example, it just outputs 0. Is there some sort of numpy function I can use to force my code to use more bits to store the number?

Hmmm, its an exponential function so I don't know if you're going to be able to solve this but adding more bits of precision. Is there not some normalisation factor that needs applying here?

sjperkins:
@JoshVStaden. Just in terms of getting a sense of scale and setting up a basic example, note the following which works on master branch:

Assuming that the PHASE_DIR in the FIELD subtable of the Measurement Set is (0, 60) degrees,  I can predict 5 point sources in a cross pattern with the following sky-model..txt:

```
# ra, dec, I, Q, U, V
00d00m00.00s, 59d59m00.00s, 1.0, 0.0, 0.0, 0.0
00d00m00.00s, 60d00m00.00s, 1.0, 0.0, 0.0, 0.0
00d00m00.00s, 60d01m00.00s, 1.0, 0.0, 0.0, 0.0
-00d01m00.00s, 60d00m00.00s, 1.0, 0.0, 0.0, 0.0
00d01m00.00s, 60d00m00.00s, 1.0, 0.0, 0.0, 0.0
```

The following commands (on master branch):

```bash
$ python predict.py -sm sky-model.txt ~/data/WSRT.MS
$ wsclean -weight uniform -name uni_coeff0 -size 512 512 -scale 1.5asec -make-psf -data-column MODEL_DATA ~/data/WSRT.MS
$ ds9  uni_coeff0-image.fits  # Substitute with tigger or some other fits viewer
```

produce the following image:

ds9

o-smirnov:
> python, if I input np.exp(-92 ** 2), for example, it just outputs 0

Well that's exactly what I suggested earlier ("Maybe somewhere the scaling of the u, v coordinates is so wrong that it tapers down to 0 immediately?"). The problem is having something as high as "92" in the exponent. Are you normalizing for the wavelength?

You need to play with some FFTs to get a feel for what's realistic and what's not:

* The FT of a unity point source (i.e. a delta function) is a constant. So you would expect a constant amplitude of 1 in the visibilities.

* The FT of a small Gaussian (only slightly bigger than a point source) is a very "wide" Gaussian in the visibilities. The amplitude will be dropping slowly towards the longer baselines.

sjperkins:
> Maybe somewhere the scaling of the u, v coordinates is so wrong that it tapers down to 0 immediately?"). The problem is having something as high as "92" in the exponent. Are you normalizing for the wavelength?

Yes, this would need to be done in a fashion similar to the gaussian shape:

https://github.com/ska-sa/codex-africanus/blob/98460bcd4ccfe33101d5b36812fc244dcb087953/africanus/model/shape/gaussian_shape.py#L47-L48

Note that this will introduce a frequency dimension into the resulting output variable

sjperkins:
> Yes, this would need to be done in a fashion similar to the gaussian shape

Note the frequency scaling factor is usually just `freq/lightspeed`. lightspeed is defined as c in `africanus.constants`

JoshVStaden:
I have done the frequency scaling. However, the scaling wasn't enough to get the values small enough, so I additionally divided the uv coordinates by 3000, which seems to be giving me that shapes we are wanting. The amplitudes taper off now, and wsclean does give me a gaussian.
uni_coeff0_image
vis_vs_baseline
I have attached the image generated by wsclean, as well as the graph of visibility amplitude vs baseline. 

@o-smirnov When you say normalizing for the wavelength, do you mean frequency scaling? Or is this something different?

sjperkins:
> When you say normalizing for the wavelength, do you mean frequency scaling? Or is this something different?

wavelength = frequency / lightspeed, so yes it is the same thing.

Its promising that we're seeing something reasonable, but it would be good to change the 3000 fudge factor something more explicable ;-)

landmanbester:
That is encouraging. Suspicious that its a factor of 3000 because
lightspeed is approx 3e8. What units are your frequencies in? Where do you
take lightspeed from?

JoshVStaden:
It seems that the frequency values are actually much larger than the gauss scale, and that's why I still need to divide. I'm not too sure what units the frequency is in. One example of the values I am getting is 1400062500. Otherwise, I can't seem to find info on what units the frequencies are in.

sjperkins:
> It seems that the frequency values are actually much larger than the gauss scale, and that's why I still need to divide. I'm not too sure what units the frequency is in. One example of the values I am getting is 1400062500. Otherwise, I can't seem to find info on what units the frequencies are in.

They're in [Hertz](https://en.wikipedia.org/wiki/Hertz), so that figure you're reading is 1.4GHz.

JoshVStaden:
Okay. Should my script be reading the frequencies in MHz or GHz instead?

sjperkins:
Standard Hertz is fine. So for example

1400062500 Hz divided 299792458 m / s == 46.66688

that's the main factor that you should be multiplying into your UV coordinate.

Some other scaling factor may be involved though and I wouldn't know what that is.

JoshVStaden:
True, but the uv coordinate is too big already. We would have to multiply by a factor much smaller than 1.

JoshVStaden:
@landmanbester @sjperkins Hi. I have started writing up the Fourier space test, which I pushed to GitHub. I am stuck with a few things. Firstly, I just wanted to check that I am doing the FFT correctly. The line of code I did for the FFT can be found [here.](https://github.com/JoshVStaden/codex-africanus/blob/e7b0b6342952d16eb1d9740bb37f7bb2628b0d7f/africanus/model/shape/tests/test_shapelets.py#L117) I have also attached an image showing the results of this code, titled FFT Shapelets:
ft_shapelets
For the FFT, I simply took the same array from image space and did the FFT. I have also attached the shapelets generated from image space, titled Image Space Shapelets:
img_shapelets
As expected, a large Gaussian in image space created a small Gaussian in Fourier space. I would just like to check that this is indeed what we expected? 

The next thing I wanted to check was if there is anything wrong with how I am doing the analytic shapelets (which I think might be the case). Just for now, I am still using [beta values of 1](https://github.com/JoshVStaden/codex-africanus/blob/e7b0b6342952d16eb1d9740bb37f7bb2628b0d7f/africanus/model/shape/tests/test_shapelets.py#L21), reason being that even a beta value of 1 creates shapelets that are too small. In order to get my script to output anything, I also gave the script [a frequency value of 1e6](https://github.com/JoshVStaden/codex-africanus/blob/e7b0b6342952d16eb1d9740bb37f7bb2628b0d7f/africanus/model/shape/tests/test_shapelets.py#L105). With this, I get the following image:
ca_shapelets
It seems that increasing the frequency value decreases the size of the Gaussian. Am I on the right track here? Is there anything that I am doing wrong that is immediately apparent?

landmanbester:
Hi Josh,

On Thu, Jun 6, 2019, 08:45 JoshVStaden <notifications@github.com> wrote:

> @landmanbester <https://github.com/landmanbester> @sjperkins
> <https://github.com/sjperkins> Hi. I have started writing up the Fourier
> space test, which I pushed to GitHub. I am stuck with a few things.
> Firstly, I just wanted to check that I am doing the FFT correctly. The line
> of code I did for the FFT can be found here.
> <https://github.com/JoshVStaden/codex-africanus/blob/e7b0b6342952d16eb1d9740bb37f7bb2628b0d7f/africanus/model/shape/tests/test_shapelets.py#L117>
>
Looks like you are computing the fftfreqs using cell size in uv space,
> shoukd be cell size in image space.
>

II have also attached an image showing the results of this code, titled FFT
> Shapelets:
> [image: ft_shapelets]
> <https://user-images.githubusercontent.com/28093948/59011761-f5344b00-8835-11e9-82cb-5693cbcbdec3.png>
> For the FFT, I simply took the same array from image space and did the
> FFT. I have also attached the shapelets generated from image space, titled
> Image Space Shapelets:
> [image: img_shapelets]
> <https://user-images.githubusercontent.com/28093948/59011830-2b71ca80-8836-11e9-8dd8-85d4b4816231.png>
> As expected, a large Gaussian in image space created a small Gaussian in
> Fourier space. I would just like to check that this is indeed what we
> expected?
>
Indeed it is.

> The next thing I wanted to check was if there is anything wrong with how I
> am doing the analytic shapelets (which I think might be the case).
>
I suspect this is due to incorrectly computing the fftfreqs

> Just for now, I am still using beta values of 1
> <https://github.com/JoshVStaden/codex-africanus/blob/e7b0b6342952d16eb1d9740bb37f7bb2628b0d7f/africanus/model/shape/tests/test_shapelets.py#L21>,
> reason being that even a beta value of 1 creates shapelets that are too
> small. In order to get my script to output anything, I also gave the script a
> frequency value of 1e6
> <https://github.com/JoshVStaden/codex-africanus/blob/e7b0b6342952d16eb1d9740bb37f7bb2628b0d7f/africanus/model/shape/tests/test_shapelets.py#L105>.
> With this, I get the following image:
> [image: ca_shapelets]
> <https://user-images.githubusercontent.com/28093948/59012180-23fef100-8837-11e9-85d6-d49a3f5fd321.png>
> It seems that increasing the frequency value decreases the size of the
> Gaussian.
>
I think it's a good exercise for you to try and explain this behaviour to
us. What does it mean for the uv coordinates if you increase the frequency?

JoshVStaden:
Okay. So the fftfreqs is what needs to be fixed. The issue is that the FFT and the analytic versions need to be the same shape. My reasoning here is to keep the same size grid between the uv and lm coordinates (therefore I divide the range of uv values by npix_x and npix_y). I thought that this would result in a different cell size between the lm and uv coordinates, due to their ranges of values being different, while keeping the array size the same. Instead of this, should I instead work out the Nyquist sampling frequency for the uv grid? Thought that only applied to the lm grid.

In terms of the frequency scaling, it makes sense that the Gaussian would get smaller. As the frequency gets higher, as each uv coordinate scales with the frequency. If the coordinates get plugged into e^(-u^2), this would have an exponential effect on the Gaussian, as this would become e^(-(1000u)^2). Because one step up in u now becomes the equivalent of 1000 steps up in the Gaussian, this would naturally cause a sharper rise in the Gaussian. Is there a specific frequency value that I should pick for my script?

landmanbester:
You shouldn't need to worry about the frequency too much as long as it doesn't cause an overflow in the exponent (remember there is a exp(-2 pi j nu/c(...)) term in the dft). When using the FFT the lm grid actually completely determines what the uv grid will be. Recall that the max frequency is going to be 1/cell_size and the min frequency is going to be 1/fov. To get the fftfreqs you could do something like

u_freqs = fftshift(np.fftfreq(npix, delta_l))
v_freqs = fftshift(np.fftfreq(npix, delta_m))
uu, vv = np.meshgrid(u_freqs, v_freqs)
uv = np.vstack(uu.flatten(), vv.flatten())

If you've done it this way then you shouldn't need the Fs on ca_shapelets that you do [here](https://github.com/JoshVStaden/codex-africanus/blob/e7b0b6342952d16eb1d9740bb37f7bb2628b0d7f/africanus/model/shape/tests/test_shapelets.py#L118). I'm not sure what you were doing there anyway because this is a Fs of a flattened 2D array. To determine the  extent of the lm coordinates you simply need to make sure that they are big enough so that the zero order shaplet fits on the grid. Hope that clarifies things a little

JoshVStaden:
Instead of delta_l and delta_m, I just lazily have a delta_x, which just describes the grid size for the lm grid. I just tried using that instead, and there seems to be no change to my output. Also, taking away the Fs causes the following to happen:
ca_shapelets

JoshVStaden:
I have noticed an additional peculiarity in my code. I tried commenting out the scaling factors in my script that make it a Fourier domain shapelet (i.e. the i^(n1 + n2) and the beta^(-1)), and comparing that to the image space shapelet, which I know is correct. I also gave my script lm coordinates just for now. As expected, the [image space shapelets](https://github.com/JoshVStaden/codex-africanus/blob/a86dcfc74acfdc5af45c2aa9d87835ce091fd2e8/africanus/model/shape/tests/test_shapelets.py#L78-L86) output exactly what we expected:
img_shapelets
The [shapelets in the script](https://github.com/JoshVStaden/codex-africanus/blob/a86dcfc74acfdc5af45c2aa9d87835ce091fd2e8/africanus/model/shape/tests/test_shapelets.py#L126-L130) output the same range of values, but it seems to have shifted the center to the top of the image:
ca_shapelets
It seems that no amount of tinkering with how my script handles the coordinates seems to change this. In fact, the only way I can get these values to match is to do a fftshift on the shapelets.

landmanbester:
It looks like you are doing a fftshift on a flattened 2D array

landmanbester:
Why are you doing the Fs [here](https://github.com/JoshVStaden/codex-africanus/blob/a86dcfc74acfdc5af45c2aa9d87835ce091fd2e8/africanus/model/shape/tests/test_shapelets.py#L141)? What happens if you remove it?

JoshVStaden:
Sorry, that seems to be a piece of code that I inserted there for some reason, meant to take out, and ended up not doing. Seems to fix this issue when I remove it, thanks for spotting it. Now that I know it is correctly looping through the coordinates, frequency, etc., and handling the coefficients properly (otherwise it wouldn't pass), I can start testing it in the Fourier domain again.

JoshVStaden:
@landmanbester Hi Landman. I got the 2D shapelets working. There seems to be a peculiarity with the imaginary part, for both the 1D and 2D case. The imaginary parts don't seem to be matching up. Is this an issue, or would it not really affect anything because we're only interested in the amplitudes?

landmanbester:
The imaginary parts should be zero for the 0th order shapelet since its symmetric. The value from the FFT could be something very close to zero especially if an even number of pixels are used

JoshVStaden:
Sorry my bad. Let me backtrack a bit. I tried seeing how the shapelets behaved when more coefficients (all 1) were added. The amplitudes always matched each other, but the phases didn't. I was just wondering if that was something that needed fixing.

landmanbester:
Yes probably. Are you using an odd or an even grid? Try using an odd grid for the tests

JoshVStaden:
I am using an odd grid. It seems that one array is the conjugate of the other (i.e. the one array's imaginary values are the negative of the other).

JoshVStaden:
@sjperkins Hi Simon. I am having a few issues with Dask. Whenever I run my script, [the blockwise function call](https://github.com/JoshVStaden/codex-africanus/blob/21b7cc46101037ec84e4fd6eb63b0036656f1980/africanus/model/shape/dask.py#L44) gives me the following error:

Traceback (most recent call last):
  File "predict_shapelet.py", line 410, in <module>
    predict(args)
  File "predict_shapelet.py", line 302, in predict
    da.from_array(lm, chunks=lm.shape))
  File "/home/vanstanden/codex-africanus/africanus/model/shape/dask.py", line 52, in shapelet
    dtype=dtype)
  File "/home/vanstanden/ca/local/lib/python2.7/site-packages/dask/array/blockwise.py", line 135, in blockwise
    chunkss, arrays = unify_chunks(*args)
  File "/home/vanstanden/ca/local/lib/python2.7/site-packages/dask/array/core.py", line 2574, in unify_chunks
    for n, j in enumerate(i))
  File "/home/vanstanden/ca/local/lib/python2.7/site-packages/dask/array/core.py", line 2574, in <genexpr>
    for n, j in enumerate(i))
IndexError: tuple index out of range

This seems to be a constant issue with calling the blockwise function. Just so we understand, this function takes in the different blocks, indexes them, and applies the shapelet function to the different blocks? What is the cause of the tuple index out of range error? In a previous version of the shapelets, I found that doing something similar to what you did [here](https://github.com/JoshVStaden/codex-africanus/blob/21b7cc46101037ec84e4fd6eb63b0036656f1980/africanus/model/shape/dask.py#L25) seemed to fix it, but didn't quite understand it.

sjperkins:
> @sjperkins Hi Simon. I am having a few issues with Dask. Whenever I run my script, [the blockwise function call](https://github.com/JoshVStaden/codex-africanus/blob/21b7cc46101037ec84e4fd6eb63b0036656f1980/africanus/model/shape/dask.py#L44) gives me the following error:
> 
> Traceback (most recent call last):
> File "predict_shapelet.py", line 410, in
> predict(args)
> File "predict_shapelet.py", line 302, in predict
> da.from_array(lm, chunks=lm.shape))
> File "/home/vanstanden/codex-africanus/africanus/model/shape/dask.py", line 52, in shapelet
> dtype=dtype)
> File "/home/vanstanden/ca/local/lib/python2.7/site-packages/dask/array/blockwise.py", line 135, in blockwise
> chunkss, arrays = unify_chunks(*args)
> File "/home/vanstanden/ca/local/lib/python2.7/site-packages/dask/array/core.py", line 2574, in unify_chunks
> for n, j in enumerate(i))
> File "/home/vanstanden/ca/local/lib/python2.7/site-packages/dask/array/core.py", line 2574, in
> for n, j in enumerate(i))
> IndexError: tuple index out of range
> 
> This seems to be a constant issue with calling the blockwise function. Just so we understand, this function takes in the different blocks, indexes them, and applies the shapelet function to the different blocks? What is the cause of the tuple index out of range error? In a previous version of the shapelets, I found that doing something similar to what you did [here](https://github.com/JoshVStaden/codex-africanus/blob/21b7cc46101037ec84e4fd6eb63b0036656f1980/africanus/model/shape/dask.py#L25) seemed to fix it, but didn't quite understand it.

Conceptually a dask array is a collection of numpy arrays (ndarray),  where each ndarray corresponds to a chunk of the dask array.

- If we had a 1D dask array, you could think of this as a list of ndarrays.
- If we had a 2D dask array, you could think of this as a list of a list of ndarrays
- The greater the ndim in the dask array, the greater the nesting of the ndarray lists.

When we call ``dask.array.blockwise`` we're mapping chunks (ndarrays) of input arrays onto chunks (ndarrays) of the output arrays.

If we call something like the following:

```python
def _block_func(ant1, ant2, vis):
    ...
    return func_result


result = da.blockwise(_block_func, ("row", "chan", "corr"),
                      ant1, ("row",),
                      ant2, ("row",),
                      vis, ("row", "chan", "corr"),
                      dtype=vis.dtype)
```

then each input chunk (for vis especiialy) maps directly to an output chunk **as an ndarray**. So the ``_block_func`` args will all receive ndarrays.

By contrast, let us drop the row dimension in the output:

```python
def _block_func(ant1, ant2, vis):
    ...
    return func_result


result = da.blockwise(_block_func, ("chan", "corr"),
                      ant1, ("row",),
                      ant2, ("row",),
                      vis, ("row", "chan", "corr"),
                      dtype=vis.dtype)
```

Now ``_block_func` will receive a **list of ndarrays** in the **vis** arg. This is because all row chunks contribute to a single chunk of the output (there is a contraction over the row dimension).

As a result, one needs to watch out for input dimensions that are missing in the output -- there'll be a contraction and the argument will be passed through as a (possibly nested) list of ndarrays.

Beware of using this naively:

```python
def _block_func(ant1, ant2, vis):
    vis = vis[0]
```

because it's assuming two things

1. A contraction over a dimension
2. That there'll only be a single chunk in the contracted dimension

A more general approach might be something as follows

```python
def _block_func(ant1, ant2, vis):
    chan, corr = ...
    res = np.empty((chan, corr), dtype=vis[0].dtype

    for v in vis:
         res += v.sum(axis=0)

    return res
```

Moral of the story

1. Think of a dask array as a nested list of ndarrays
2. blockwise maps input chunks to output chunks
3. contractions of input dimensions mean that nested lists of ndarrays will be passed into ``_block_func``, with the level of nesting corresponding to the number of dimensions contracted.

sjperkins:
One other thing I forgot to mention.

If the blockwise output is 2D. i.e. ("chan", "corr"), then the result  returned by ``_block_func`` should also be 2D, otherwise dask will have issues piecing the parts of the array together when compute is called on it.

sjperkins:
So looking at your blockwise call:

```python
    return da.blockwise(_shapelet_wrapper, ("row", "chan", "source" ),
                        coords, ("row", "coord-comp"),
                        frequency, ("chan",),
                        coeffs_l, ("source", "nmax1"),
                        coeffs_m, ("source", "nmax2"),
                        beta, ("source", "beta-comp"),
                        delta_lm, ("delta_lm-comp",),
                        lm, ("source", "lm-comp"),
```

All your input arrays (except frequency) have a contraction over one dimension. And since ``coord-comp``, ``nmax1``, ``nmax2``, ``beta-comp``, ``delta_lm-comp`` and ``lm-comp`` probably only have a single chunk in them, you could do this trick:

```python
def fn(arg1, arg2):
   arg1 = arg1[0]  # Reference the single chunk in the contraction
```

I would assert that those dimensions have a single chunk in the function calling blockwise though.

JoshVStaden:
@landmanbester Hi Landman. I have created a test case for my shapelets to test them against the Gaussian code. One question, I see that the Gaussian shape code doesn't have something to handle the w term (as mine does). How did we say that we were going to test the w term again?

In terms of imaging the shapelet in the predict script with wsclean, the script runs within an acceptable time frame (thanks Simon for spotting my mistake), however I am encountering the same issues I had a while ago, where nothing would be imaged due to frequency values being too high. It seems to generate an image if I divide frequency by 1000, though. It seems that the shapelets themselves seem to generate non-zero values, and the same with the predict_vis script. How did we say we were going to fix this again?

landmanbester:
You need to test your zeroth-order shapelet against the Gaussian implementation in codex. What do you mean there is no w-term? Which function are you looking at?

You shouldn't have to divide the frequency by anything but the speed of light. Keep in mind that beta will have to correspond to something physical. What are your beta values and lm coordinates looking like?

JoshVStaden:
I can't see any w-term handling in the script that I am using. I am using [the gaussian_shape.py script used in model/shape.](https://github.com/JoshVStaden/codex-africanus/blob/shapelets/africanus/model/shape/gaussian_shape.py) I can see that the u, v, and w coordinates are extracted [here](https://github.com/JoshVStaden/codex-africanus/blob/e75cded02a55fddd656c7095179814375be5af0f/africanus/model/shape/gaussian_shape.py#L41), but then the w coordinate is not handled. 

So, currently, I am running the test script on a regular grid to test against the Gaussian, and then I am running the predict script to try to image the shapelets. In order to test against the Gaussian, I have kept beta as 1, and manipulated the inputs so that I can simply get the e^(-(u^2 + v^2)) (as can be seen in the gaussian script [here](https://github.com/JoshVStaden/codex-africanus/blob/e75cded02a55fddd656c7095179814375be5af0f/africanus/model/shape/gaussian_shape.py#L50)).

For the predict script, I have realized while reading your feedback that I made the beta values too large. They were 0.01 when testing on Sunday, and I have simply reduced them to 1e-7, and now I get the following image:

zero_coeff_shapelet

which seems to be correct. So at least the imaging works. Now it is just the testing against the Gaussian that needs to work.

sjperkins:
@JoshVStaden https://github.com/ska-sa/codex-africanus/pull/129 has now updated the predict example to support Tigger LSM files as well as the Gaussian's described by them.

Taking a look at the first four lines of the ASCII format:
```
#format: name ra_d dec_d i spi freq0 emaj_s emin_s pa_d
J0 0.295293 -30.333156 0.013291 -0.700000 1420000000.000000 8.444299 7.789684 29.186436
J1 0.282713 -29.820104 0.013647 -0.700000 1420000000.000000 0.000000 0.000000 75.540160
J2 -0.328503 -29.665906 0.017758 -0.700000 1420000000.000000 0.000000 0.000000 6.673732
```

Each line describes a source, with the following columns:

name: Useful Source Name
ra_d: Right Ascension in degrees
dec_d: Declination in degrees
i: I stokes value
spi: Spectral index
freq0: Reference Frequency
emaj_s: Gaussian major axis (in arcseconds I think @o-smirnov?)
emin_s: Gaussian minor axis
pa_d: Positiion angle in degrees

sjperkins:
An emaj_s and emin_s of 0 implies that the source is a point source.

JoshVStaden:
Hi. Just a quick update, I have created [my own branch for Tigger-lsm.](https://github.com/JoshVStaden/tigger-lsm/tree/shapelet_model) So far, I only have support for a single shapelet, and just need to extend it to have multiple coefficients. The format for shapelets in the lsm file is as follows:
`name ra_d dec_d i spi freq0 sbetal_s sbetam_s scoeffsl scoeffsm`
When running it with the predict script, I get the following image (same as for the Gaussian)
tigger
 
Is my file format correct, or are there any modifications that should be done to it? And does the image look right to you?

@sjperkins Should we set up a meeting tomorrow to discuss this?

sjperkins:
> Is my file format correct, or are there any modifications that should be done to it? 

Not sure about the file format, this might be something that you'd like to define yourself!

> And does the image look right to you?

No, but I suspect that's because the source coordinates (ra_d, dec_d) are wrong for the phase centre in the MS... I forgot that the predict example was to produce 5 star's around phase centre. So I think you'll have t o modify them appropriately.

However, I am confident that the predict script does the right thing (TM) because it matches up with MeqTrees.

> @sjperkins Should we set up a meeting tomorrow to discuss this?

Yes, lets hangouts at the usual time.

JoshVStaden:
@landmanbester @sjperkins Hi. I have put in a script to randomly generate shapelets. There is one main issue with the script so far: if the sky model has a set of shapelets with an equal number of coefficients describing all of them, it returns a single array of shape (nsrc, nmax1, nmax2). If, however, there are sources with varying number of coefficients describing them, then it will return an array of shape (nsrc,), and each element in the array is of object type, which isn't supported in Numba.

sjperkins:
@JoshVStaden how about making the number of coefficients the same, but having some marker coefficient such as zero or -1 that results in the coefficient being ignored?

JoshVStaden:
@sjperkins That could be a good idea for now. I can just put a check in the shapelet script that skips any coefficients that are zero. Makes sense to do that in the first place.

landmanbester:
Well if the coefficient is zero it won't add anything anyway but computing it is obviously a waste. Just to be clear, what does your format line look like for the lsm? It should be something like

#format: name ra dec i q u v spi freq0 beta_l beta_m maxorder coeffs  

where coeffs is the list of coefficients as discussed and maxorder is the maximum order of the 1D basis functions and we are assuming it is the same for both l and m. Is that what you are doing?

JoshVStaden:
I still have yet to update my format to include that. I left a comment on the tigger-lsm pull request asking how the maxorder field would be implemented, and suggesting another possible way of handling the coeffs field.

Just to confirm, say coeffs has the value `1,2,3,4,5,6,7,8,9` and maxorder is 3, would this give the coefficients matrix a value of 
```
1 3 6 
2 5 8
4 7 9
```

landmanbester:
Yep, I think that is fine. At the end of the day, you have to decide on a convention, document it, and then stick to it

o-smirnov:
OK hang on, if I look at the literature, the "order" of a shapelet is max of N1+N2. See e.g. Fig. 3 of  https://arxiv.org/pdf/astro-ph/0105178.pdf

So to me, if you have a shapelet decomposition of order 2, then you only have non-zero coefficients given by 

```
c00 c01 c02
c10 c11 
c20
```

and then the ordering in the coefficient list is

```
0 2 5
1 4 
3
```

and then you don't actually need a ``max_order`` parameter at all -- you just look at the lenght of coefficient list you've been given. Clear? Just discussed it with @landmanbester, and I think we're on the same page. Forget the square -- use the triangle.

JoshVStaden:
@o-smirnov I have this implemented. So the official LSM format for shapelets is 

`
#format: name ra_d dec_d i spi freq0 sbetal_s sbetam_s shapelet_coeffs
`

and the shapelet_coeffs will be read in as a triangle.

JoshVStaden:
@sjperkins @landmanbester Hi. I have added the primary beam into the script. I have attached the image I obtained when I added it in. 

primary_beam

The issue is that there does not seem to be any primary beam showing. I just wanted to check if I've passed my lm coordinates in correctly. [I got the radec coordinates from the sky model for each source and passed them in](https://github.com/JoshVStaden/codex-africanus/blob/ebacd6a00e5c193a9edd4ecd0b8a4450748b28a2/africanus/rime/examples/predict_shapelet.py#L354). Was this the correct thing to do?

landmanbester:
Well, I do see a slightly circular pattern in the image. Maybe your image is just a bit small. Can you produce an image of the absolute value of the primary beam for the same field of view?

JoshVStaden:
zernike_primary_beam
Think I might be seeing the issue here...

JoshVStaden:
zernike_primary_beam
Oh wait, I've made an error in generating the primary beam. Here is the primary beam.

sjperkins:
I went through the montblanc and codex-africanus beam code and it looks like we're doing the rightish (i.e. acceptable) thing by passing lm coordinates into the beam (as opposed to radec).

So the parangle rotation in the numba beam is right  (especially since it agrees with Montblanc. The ddfacet branch if memory serves)

https://github.com/ska-sa/codex-africanus/blob/ab32d111900058318e963ead439b09f73e3db171/africanus/rime/tests/test_fast_beams.py#L210

sjperkins:
> I went through the montblanc and codex-africanus beam code and it looks like we're doing the rightish (i.e. acceptable) thing by passing lm coordinates into the beam (as opposed to radec).
> 
> So the parangle rotation in the numba beam is right (especially since it agrees with Montblanc. The ddfacet branch if memory serves)
> 
> https://github.com/ska-sa/codex-africanus/blob/ab32d111900058318e963ead439b09f73e3db171/africanus/rime/tests/test_fast_beams.py#L210

Any comments here @bennahugo?

sjperkins:
Parallactic angles are computed as follows:

https://codex-africanus.readthedocs.io/en/latest/rime-api.html#africanus.rime.parallactic_angles

- antenna positions are POSITION in the ANTENNA sub-table
- times are unique times.


Then, you'll need to implement the following transforms in zernike_dde: 

- https://github.com/ska-sa/codex-africanus/blob/a9c53c4f88d8a8b93987a429c60b9908d683f2c6/africanus/rime/fast_beam_cubes.py#L127-L151

JoshVStaden:
@sjperkins Hi Simon. I managed to find the antenna positions, but am having issues with the time indices. I can't seem to find the timestamps in the measurement set. I understand that there is a TIME table, and that finding the max of the time_index will give ntime, but I don't understand why time_index has a shape of (7560,) and ntime is 72. I assume that I would use time_index as a way of indexing ms.TIME.data to find the actual timestamps, but how do I use time_index to find the actual timestamps?

sjperkins:
> @sjperkins Hi Simon. I managed to find the antenna positions, but am having issues with the time indices. I can't seem to find the timestamps in the measurement set. I understand that there is a TIME table, and that finding the max of the time_index will give ntime, but I don't understand why time_index has a shape of (7560,) and ntime is 72. I assume that I would use time_index as a way of indexing ms.TIME.data to find the actual timestamps, but how do I use time_index to find the actual timestamps?

This is slightly complicated if you're dealing with dask arrays, but this is a useful pattern to learn:

```python
from operator import getitem
import numpy as np
time = ds[0].TIME.data

# Call np.unique on each chunk (block) of the time dask array
# np.unique returns a (utime, time_index) tuple
# so each chunk of the resulting array is a tuple
intermediate = time.map_blocks(lambda t: np.unique(t, return_inverse=True),
                               meta=np.empty((0,), dtype=tuple),  # dask needs meta for py3
                               dtype=tuple)
 
# Create a dask array where each chunk corresponds to the unique times
# in the first tuple element of each chunk in intermediate
#  The chunk size is set to nan becaues we don't know how many
# unique times are in each row. Or you could precompute the chunk sizes...
utime = intermediate.map_blocks(lambda i: getitem(i, 0),
                              chunks=(np.nan,),
                              dtype=time.dtype)

# Create a dask array where each chunk corresponds to the time indices
# in the second tuple element of each chunk in intermediate.
# The chunk sizes will match those of the original array
time_index = intermediate.map_blocks(lambda i: getitem(i, 1),
                               dtype=np.int32)

```

JoshVStaden:
Thanks Simon. Seems to be reading the times fine. I do, however, seem to be having an issue where my code is throwing the following error:

```SEVERE  MeasTable::dUTC(Double) (file /build/casacore-3.1.1/measures/Measures/MeasTable.cc, line 4281)  Cannot read leap second table TAI_UTC```

when calling

```pa = parallactic_angles(utime, da.from_array(antenna_positions), da.from_array(field.PHASE_DIR.data)).compute()```

I have made sure that each input is a dask array of correct shape. Any potential reason why this might be happening?

**Comment by sjperkins** [comment]
> Thanks Simon. Seems to be reading the times fine. I do, however, seem to be having an issue where my code is throwing the following error:
> 
> `SEVERE MeasTable::dUTC(Double) (file /build/casacore-3.1.1/measures/Measures/MeasTable.cc, line 4281) Cannot read leap second table TAI_UTC`
> 
> when calling
> 
> `pa = parallactic_angles(utime, da.from_array(antenna_positions), da.from_array(field.PHASE_DIR.data)).compute()`
> 
> I have made sure that each input is a dask array of correct shape. Any potential reason why this might be happening?

This could possibly be because your the kernsuite casacore-data package is out of date. It contains astronomical measurements which are frequently updated. If you're on an ubuntu/debian system this should be as simple as:

```bash
$ sudo apt update
$ sudo apt upgrade casacore-data
```

Otherwise, if you've got python-casacore 3.1.1 installed, this might be the issue: https://github.com/casacore/python-casacore/issues/181, in which case you'll need to set an environment variable to point it at the right measures data.

JoshVStaden:
I am running this on Jake. It seems that python-casacore is up to date at least, so it must be the referenced issue. I think I might need to set the environment variable, but I'm not sure which environment variable to set to which value.

sjperkins:
The debian python-casacore package on jake is python-casacore 3.0.0.

```bash
sjperkins@jake:~$ dpkg -l python-casacore
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                                          Version                     Architecture                Description
+++-=============================================-===========================-===========================-===============================================================================================
ii  python-casacore                               3.0.0-1kern1                amd64                       Python 2 bindings to the casacore library
```

Note that if you're operating in a python virtual environment you may have a python-casacore package different from the debian package.

sjperkins:
Apologies I was wrong about the environment variable. You'll need to set something in a ``~/.casarc`` file. For more information please reread the python-casacore issue.

JoshVStaden:
@landmanbester @sjperkins Hi. Changes have been made to incorporate the parallactic angle rotation, frequency scaling, antenna scaling, and pointing error. I am still keeping frequency scaling and antenna scaling to 1, and pointing error to 0. With these changes made, if any difference has been made because of it, it is too small for me to notice. I have attached two images, the first one being the image without a primary beam, and the second one with a primary beam. Does this look the same as previously, where we had multiple primary beams centered on each source?
no_pb
with_pb

landmanbester:
I can still see a weird circular pattern but not around each source. How big is the field of view in your simulation? Are the sources even close to the edge of the beam? The beam effect might not be obvious otherwise

JoshVStaden:
According to the units used in the sky model, the sources in this simulation should be varying by about 1/5 of a degree. Would this not be too big to even see a primary beam?

landmanbester:
I'm not sure what you mean by that. You mean the extent of the sources should be about 0.2 deg? It's not really about the size of the source but rather where they are relative to the centre of the beam. The beam should be about 3deg across and is reasonably flat in the middle so you won't really notice it there, especially if you have normalised it to one in the centre. So you have to make sure you have a large enough field of view for some of the sources to get close to the edge. If you look carefully you will see that there is some attenuation of the sources, especially for the one in the bottom left corner. Should be clearer if you spread the sources a bit more and give the figure a colorbar

JoshVStaden:
I meant that the sources didn't vary further than 0.2 degrees from phase center. I have altered the sky model so that field of view is 3 degrees each way (total of 6 degrees for ra, and 6 degrees for dec). I don't know how to give the figure a colorbar, as I am using wsclean to image the sources. Is there a colorbar option? I have attached the new images, again the first one has no primary beam and the second one does.
no_pb
with_pb

sjperkins:
FYI I'm working on adding holographic beam cube support to the predict example in https://github.com/ska-sa/codex-africanus/pull/145, which should hopefully help in tracking this down.

JoshVStaden:
Command being used is ```wsclean -name multiple_sources_pb -size 2048 2048 -scale 12.0asec -make-psf -data-column MODEL_DATA WSRT.MS/```

JoshVStaden:
@landmanbester Hi. I was working on generating a fits primary beam image, and I noticed something odd. It seems that the primary beam is inverting itself (i.e. the lowest value is in the center, and it seems that the highest value is 0). I have also noticed that it is not perfectly circular anymore. I am currently working on finding out why that is the case. I'll let you know if something comes up.
primary_beam

JoshVStaden:
Oh, and in terms of coordinates, this is for l and m coordinates between -1 and 1. This should mean that the beam has a radius of 1 degree, correct?

landmanbester:
No l and m can't vary by that much. Recall n = sqrt(1 - l**2 - m**2) >= 1, they are direction cosines on the sphere. I think you are confused because the beam is fit in a unit disk i.e. |rho| < 1. That's why you have to get the phyiscal coordinates by using eidos to make you a fits beam. Somewhere Khan is taking the holography data and rescaling coordinates so that he can fit in the unit disk. I'm not sure what could be causing the beam to invert. Maybe you have a convention reversed somewhere. Have you tried to read Khan's beam paper?

JoshVStaden:
Okay, I did think it was because the beam was constrained to a unit disk.

Okay, so, in essence, I get eidos to generate a fits file, which will contain coordinates. Do I then just plug these coordinates into my Zernike script and see what it produces?

Yea, the inversion might have happened somewhere while doing the parallactic angles and such. And yes, I have read Khans beam paper.

landmanbester:
Yes you should take the coordinates from the fits file and verify that you get the same image with your Zernike script for evaluating the beam in image space

JoshVStaden:
Great, thanks for clarifying.

sjperkins:
The following, soon to be merged into master, may also be helpful for understanding:

https://github.com/ska-sa/codex-africanus/pull/145/files#diff-c3333ceed46cff64a7f197a07b5f7891

JoshVStaden:
@landmanbester Hi Landman. I have tested my Zernike polynomials against the Jones images from Eidos. It seems that there is some difference in the off-diagonal terms, the difference being that my image seems to be the Eidos image reflected along the x=y axis. I have attached the two images I obtained for the xy correlation.

Eidos_image_xy
Codex_Image_xy

In the meanwhile, in the predict script, I ran the predict function with a brightness matrix of 1 everywhere:

``` 
predict_vis(time_index, ms.ANTENNA1.data, ms.ANTENNA2.data,
                       dde, da.from_array(np.ones(jones.shape), chunks=jones.shape), dde, None, None, None) 
```

and I got the following image using wsclean with a scale of 1.5asec:

image

This should represent the primary beam as it is in the predict script. It seems to be quite small, hence it doesn't seem to be having an effect on the data. It seems that the diameter of the beam in Eidos was changed by manipulating Npix, which cannot be done in the predict script.

landmanbester:
If I had to guess, the flipped axis is probably coming from the fact that fits files are fortran contiguous. Try plotting them again after doing np.ascontiguousarray on the data from the fits file. As for your image, a visibility of all ones corresponds to a 1Jy source at the phase center so what you are showing there is actually the point spread function and not the primary beam. You had the two confused in your writeup as well so make sure you understand the difference. You will not see the effect of the primary beam on a source at the phase center since the beam is normalised to unity there. Try moving the source to the half power point of the beam and check that its flux is reduced by approximately 50%.

sjperkins:
> If I had to guess, the flipped axis is probably coming from the fact that fits files are fortran contiguous. 

Yes, take a look at this FORTRAN order handling code for the predict example:

https://github.com/ska-sa/codex-africanus/blob/a1356df4efa0dd216a0fe11c8a0452d1ee863e7a/africanus/rime/examples/predict.py#L193-L199

JoshVStaden:
@landmanbester @o-smirnov Hi. As per our discussion yesterday, I have found out where the half-power beamwidth occurs in the beam. It seems to occur somewhere upwards of 80 and sometimes 90 degrees from the center of the beam, which might explain a lot.

o-smirnov:
Well you definitely have a scaling/coordinates issue then. It should be more in the ballpark of 1 degree.

sjperkins:
@JoshVStaden  I've resolved the merge conflicts.

sjperkins:
Closed in favour of #231

PR #113, opened 2019-06-03
Calibration utils

landmanbester:
This should be ready to go. Only DIAG_DIAG option is implemented for the phase_only_GN and I need to add a test for the direction dependent case but need to move on to the gridder for the time being


africanus/calibration/tests/test_utils.py
```diff
@@ -0,0 +1,166 @@
...
+from __future__ import division
+from __future__ import print_function
+
+import numpy as np
+from numpy.testing import assert_array_almost_equal
+from africanus.dft import im_to_vis
+from africanus.averaging.support import unique_time
+from africanus.calibration.utils import residual_vis, correct_vis
+from africanus.rime.predict import predict_vis
+np.random.seed(42)
```
sjperkins: It would be better if the seed was set at the start of each test

africanus/calibration/tests/test_utils.py
```diff
@@ -0,0 +1,166 @@
...
+    model_tmp = np.transpose(model_unsubtracted, [2, 0, 1, 3])
+    time_index = np.unique(time, return_inverse=True)[1]
+    vis_unsubtracted = predict_vis(
+        time_index, ant1, ant2,
+        dde1_jones=jones_tmp,
+        source_coh=model_tmp,
+        dde2_jones=jones_tmp)
+    # residual should now be equal to unsubtracted vis
+    assert_array_almost_equal(residual, vis_unsubtracted, decimal=5)
+    return
```
sjperkins: the return could be removed

africanus/linalg/kronecker_tools.py
```diff
@@ -0,0 +1,42 @@
...
+
+import numpy as np
+
+
+def kron_matvec(A, b):
+    """
+    Computes matrix vector product of kronecker matrix in linear time.
+    :param A: an array of arrays holding matrices [K0, K1, K2, ...]
+    :param b: the RHS vector
+    :return: A.dot(b)
```
sjperkins: Could you change this to numpydoc format?

africanus/linalg/kronecker_tools.py
```diff
@@ -0,0 +1,42 @@
...
+        Z = np.einsum("ab -> ba", Z)
+        x = Z.flatten()
+    return x
+
+
+def kron_cholesky(A):
+    """
+    Computes the cholesky decomposition of a kronecker matrix
+    :param A: an array of arrays holding matrices [K1, K2, K3, ...]
+    :return:
```
sjperkins: numpydoc format would be good here too

africanus/gps/__init__.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
```
sjperkins: I think the top level __init__.py got cut and pasted here. Rather just leave it empty.

africanus/calibration/utils.py
```diff
@@ -0,0 +1,332 @@
...
+
+Parameters
+----------
+time_indices : $(array_type)
+    Time values of shape :code:`(row,)`
+antenna1 : $(array_type)
+    First antenna indices of shape :code:`(row,)`.
+antenna2 : $(array_type)
+    Second antenna indices of shape :code:`(row,)`
+counts : $(array_type)
```
sjperkins: I think change the variable name to `time_counts` and move it just after `time_indices` so the time book-keeping variables are grouped together.

africanus/calibration/utils.py
```diff
@@ -0,0 +1,332 @@
...
+    Counts of unique times in time
```
sjperkins:
```suggestion
    Counts of unique times in time of shape :code:`(utime,)`
```

africanus/calibration/utils.py
```diff
@@ -0,0 +1,332 @@
...
+    with shape :code:`(row,)`.
+antenna1 : $(array_type)
+    Antenna 1 index used to look up the antenna Jones
+    for a particular baseline
+    with shape :code:`(row,)`.
+antenna2 : $(array_type)
+    Antenna 2 index used to look up the antenna Jones
+    for a particular baseline
+    with shape :code:`(row,)`.
+counts : $(array_type)
```
sjperkins: Rename `time_counts` and move to after `time_index` here too.

africanus/calibration/phase_only.py
```diff
@@ -0,0 +1,222 @@
...
+                                        s] += (np.conj(tmp_out_array) *
+                                               tmp_out_array).real
+                                    jhr[t, ant, nu,
+                                        s] += (np.conj(tmp_out_array) *
+                                               residual[row, nu])
+        return jhj, jhr
+    return _jhj_and_jhr_fn
+
+
+def phase_only_GN(time_indices, antenna1, antenna2,
```
sjperkins: What does GN stand for? Gains? Gauss-Newton? Expanding the acronym in the function name might make it immediately understandable.

africanus/calibration/utils.py
```diff
@@ -0,0 +1,332 @@
...
+    jones_inverse_mul = jones_inverse_mul_factory(mode)
+
+    @wraps(correct_vis)
+    def _correct_vis_fn(time_indices, antenna1, antenna2,
+                        counts, jones, vis, flag):
+        jones_shape = np.shape(jones)
+        n_tim = jones_shape[0]
+        n_dir = jones_shape[3]
+        if n_dir > 1:
+            raise RuntimeError("Jones has n_dir > 1.\
```
sjperkins: I think this is a ValueError rather than a RuntimeError

africanus/calibration/phase_only.py
```diff
@@ -0,0 +1,222 @@
...
+                for row in ind:
+                    if antenna1[row] == ant or antenna2[row] == ant:
+                        p = antenna1[row]
+                        q = antenna2[row]
+                        if ant == p:
+                            sign = 1.0j
+                        elif ant == q:
+                            sign = -1.0j
+                        else:
+                            raise RuntimeError(
```
sjperkins: ValueError rather than a RuntimeError

africanus/calibration/tests/test_phase_only.py
```diff
@@ -0,0 +1,136 @@
...
+from __future__ import division
+from __future__ import print_function
+
+import numpy as np
+from numpy.testing import assert_array_almost_equal
+from africanus.dft import im_to_vis
+from africanus.averaging.support import unique_time
+from africanus.calibration import phase_only_GN
+from africanus.rime.predict import predict_vis
+np.random.seed(42)
```
sjperkins: Rather initialise the seed in the test functions themselves

africanus/calibration/tests/test_phase_only.py
```diff
@@ -0,0 +1,136 @@
...
+    data_dict["WEIGHT_SPECTRUM"] = weights
+    data_dict["TIME"] = time
+    data_dict["ANTENNA1"] = antenna1
+    data_dict["ANTENNA2"] = antenna2
+    data_dict["FLAG"] = flag
+    data_dict['JONES'] = jones
+    return data_dict
+
+
+def test_phase_only_diag_diag():
```
sjperkins: Could you parametrise this test on the different correlation permutations: ``(1,), (2,), (2,2)``

africanus/calibration/utils.py
```diff
@@ -0,0 +1,332 @@
...
+import numpy as np
+from functools import wraps
+from africanus.util.docs import DocstringTemplate
+from africanus.util.numba import generated_jit, njit
+
+DIAG_DIAG = 0
+DIAG = 1
+FULL = 2
+
+# @numba.jit(nopython=True, nogil=True, cache=True)
```
sjperkins: Remove the comment

africanus/calibration/utils.py
```diff
@@ -0,0 +1,332 @@
...
+            t4 = a11*blj[1, 1]
+            out[1, 0] = t1*b00 +\
+                t2*b00 +\
+                t3*b10 +\
+                t4*b10
+            out[1, 1] = t1*b01 +\
+                t2*b01 +\
+                t3*b11 +\
+                t4*b11
+    # return jones_inverse_mul
```
sjperkins: Remove the comment

africanus/gps/utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+import numpy as np
+
+
+def abs_diff(x, xp):
+    """
+    Gets vectorised differences between x and xp
+    :param x: NxD array of floats (inputs1)
+    :param xp: NpxD array of floats (inputs2)
+    """
```
sjperkins: Could you change this to numpydoc format?

africanus/calibration/tests/test_phase_only.py
landmanbester: DIAG_DIAG means it's the (1,) or (2,) case for both the Jones terms and the visibilities. This only tests DIAG_DIAG hence the name. I will have to parametrise on the three scenarios described in the calibration api but currently only DIAG_DIAG is implemented and an error is raised if either of the other two cases are received as inputs

landmanbester: As discussed I could parametrise the tests for correct_vis and residual_vis this way but this will require quite a bit of work. Currently only testing the DIAG_DIAG case for everything, will create an issue and implement the other tests when I get the time

africanus/calibration/phase_only.py
landmanbester: Changed it to phase_only_gauss_newton

PR #116, opened 2019-06-06
Nifty gridder support

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```


sjperkins:
@landmanbester Let me know how you want this restructured

landmanbester:
Ok no that's fine. If you are putting it in gridding/nifty there won't be any clashes

mreineck:
Just a small tip for saving memory: as long as you don't do any fancy w-stacking or similar you can use `vis2grid` and `grid2dirty` instead of `vis2grid_c` and `grid2dirty_c`. This halves the memory of the `grid` arrays and makes the FFT faster. The idea is that one can rearrange the information in the complex `grid` array in such a fashion that it fits into a real-vallued array, because of Hermitian symmetry.

You will also notice that `vis2grid_c` and `grid2dirty_c` result in a dirty image with half the amplitude of that generated by `vis2grid` and `grid2dirty`. This is because in the former I don't take the "mirrored" visibilities into account.

sjperkins:
> Just a small tip for saving memory: as long as you don't do any fancy w-stacking or similar you can use vis2grid and grid2dirty instead of vis2grid_c and grid2dirty_c. This halves the memory of the grid arrays and makes the FFT faster. The idea is that one can rearrange the information in the complex grid array in such a fashion that it fits into a real-valued array, because of Hermitian symmetry.

Cool. The code is very sub-optimal memory-wise at the moment (a grid per visibility chunk!), but I just wanted to get something up and running. Memory budgets are something that I have to consider carefully dask-wise in a general sense -- @landmanbester and I had a long conversation about it yesterday. Essentially we'll probably aim for a single linear chain of gridding operations rather than the current parallel reduction. We can then depend on the internal OpenMP threading. I reckon we still might have to deal with thread over-subscription at some point but will cross that bridge when we get to it.

> You will also notice that vis2grid_c and grid2dirty_c result in a dirty image with half the amplitude of that generated by vis2grid and grid2dirty. This is because in the former I don't take the "mirrored" visibilities into account.

Thanks, I'll keep this in mind

mreineck:
> Cool. The code is very sub-optimal memory-wise at the moment (a grid per visibility chunk!), but I just wanted to get something up and running.

If there is a way to lock the grid against concurrent write accesses (what we currently achieve via `#omp critical`) in the dask context, it's not a big problem to have just one grid for all chunks being processed on one node. You just need to replace the locking mechanism.

mreineck:
For degridding, you can of course always use just one grid.

sjperkins:
> If there is a way to lock the grid against concurrent write accesses (what we currently achieve via #omp critical) in the dask context, it's not a big problem to have just one grid for all chunks being processed on one node. You just need to replace the locking mechanism.


Well instead of a fan-pattern of grids for each chunk:

```
a     b  c   d   e  f    g  h
  \  /    \ /     \ /    \ /
    \     /         \    /
     \  /             \ /
       \              /
         \           /
           \        /
             \     /
               \ /       
```

My intention is to feed the grid through as follows which should remove the need for a lock.

```
a - b - c - d - e - f - g - h -
```

The next level of the above step is to support a configurable number of grids (say two or three) on which we wish to accumulate visibilities in parallel. In this case, there'll be two or three linear chains that are summed together in a parallel reduction (once they've reached their end).

sjperkins:
> For degridding, you can of course always use just one grid.

Indeed, this is much easier to support than the gridder (for our use-cases at least!)

mreineck:
One more thing ... the dirty images you currently get out of the gridder are complex-valued, with significant imaginary parts. I think they need to be converted to real type via `.real` instead of `np.abs()` as it was done in the `nifty_dask.py` demo.

africanus/gridding/nifty/dask.py
```diff
@@ -347,7 +347,7 @@ def grid(vis, uvw, flags, weights, frequencies, grid_config,
 
 def _nifty_dirty(grid, grid_config):
     """ Wrapper function for creating a dirty image """
-    grids = [grid_config.grid2dirty_c(grid[:, :, c])
+    grids = [grid_config.grid2dirty(grid[:, :, c].real)
```
mreineck: I think this is a misunderstanding (but I'm not absolutely sure).
If you are using the `_c` functions everywhere else, you should be calling `grid2dirty_c` here and take the real part of the result.

sjperkins: Yes I think you're right, I've fixed this.

PR #118, opened 2019-06-17
dask predict_vis streams

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```


sjperkins:
@paoloserra and @o-smirnov.

OK, I've got the constrained reduction over the source dimension of source coherencies working.

The test case has 3 row chunks. Firstly, a naive parallel reduction has many tree leaves:

fan-red-1

And here is the graph of the constrained reduction (one stream per chunk):

stream-red-1

And here is the graph of the constrained reduction (two stream per chunk):

stream-red-2

Now to see if this improves the memory footprint.

PR #127, opened 2019-07-19
Calibration utils

landmanbester:
Adds a corrupt_vis function and parametrised tests over calibration mode i.e. DIAG_DIAG, DIAG and FULL


africanus/calibration/tests/test_utils.py
```diff
@@ -26,65 +21,83 @@ def give_lm(n_dir):
...
+corr_shape_parametrization = pytest.mark.parametrize(
+    'corr_shape, jones_shape',
+    [((2,), (2,)),  # DIAG_DIAG
+     ((2, 2), (2,)),  # DIAG
+     ((2, 2), (2, 2)),  # FULL
+     ])
 
-def make_dual_pol_data(sigma_n, n_dir, sigma_f):
+
+def make_data(sigma_n, sigma_f, n_time, n_chan, n_ant,
```
sjperkins: This could be a pytest factory fixture http://doc.pytest.org/en/latest/fixture.html living in a conftest.py. See for example the [xarray-ms ms fixture](https://github.com/ska-sa/xarray-ms/blob/master/xarrayms/tests/conftest.py)

```python
@pytest.fixture
def calibration_factory():
  def impl(sigma_n, sigma_f, n_time, n_chan, n_ant, ...):
      data_dict = {}
      # Fill in data_dict
      ...
      return data_dict

  return impl
```

def test_calibration(calibration_facttory):
   data_dict = calibration_factory(sigma_n, sigma_f, n_time, ...)
```
```

africanus/calibration/tests/test_utils.py
```diff
@@ -26,65 +21,83 @@ def give_lm(n_dir):
...
     # make up some sources
     lm = give_lm(n_dir)
-    flux = give_flux(n_dir)
-    # simulate model data (pure Stokes I)
+    alpha = -0.7
+    freq0 = freq[n_chan//2]
+    flux = give_flux(n_dir, n_chan, corr_shape, alpha, freq, freq0)
+    # simulate model data
     for dir in range(n_dir):
         this_lm = lm[dir].reshape(1, 2)
```
sjperkins: How about?

```suggestion
        dir_lm = lm[dir].reshape(1, 2)
```

africanus/calibration/tests/test_utils.py
```diff
@@ -26,65 +21,83 @@ def give_lm(n_dir):
...
-        this_flux = np.tile(flux[dir], (n_chan, n_cor))[None, :, :]
-        model_tmp = im_to_vis(this_flux, uvw, this_lm, freq)
-        model_data[:, :, dir, :] = model_tmp
+        # Get flux for source (keep source axis, flatten cor axis)
+        this_flux = flux[dir].reshape(1, n_chan, np.prod(corr_shape))
```
sjperkins: How about?

```suggestion
        dir_flux = flux[dir].reshape(1, n_chan, np.prod(corr_shape))
```

africanus/calibration/tests/test_utils.py
```diff
@@ -26,65 +21,83 @@ def give_lm(n_dir):
...
+        tmp = im_to_vis(this_flux, uvw, this_lm, freq)
+        model_data[:, :, dir] = tmp.reshape((n_row, n_chan) + corr_shape)
     assert not np.isnan(model_data).any()
     # simulate gains (just radnomly scattered around 1 for now)
```
sjperkins:
```suggestion
    # simulate gains (just randomly scattered around 1 for now)
```

africanus/calibration/tests/test_utils.py
```diff
@@ -5,19 +5,14 @@
...
 
-n_time = 32
-n_chan = 16
-n_ant = 7
-n_bl = n_ant*(n_ant-1)//2
-n_row = n_bl*n_time
-n_cor = 2
-
 
 def give_lm(n_dir):
```
sjperkins: This is a https://en.wikipedia.org/wiki/Factory_method_pattern

```suggestion
def lm_factory(n_dir):
```

africanus/calibration/tests/test_utils.py
```diff
@@ -26,65 +21,83 @@ def give_lm(n_dir):
     return lm
 
 
-def give_flux(n_dir):
-    return 0.1 + np.random.random(n_dir)
+def give_flux(n_dir, n_chan, cor_shape, alpha, freq, freq0):
```
sjperkins: This is a https://en.wikipedia.org/wiki/Factory_method_pattern

```suggestion
def flux_factory(n_dir, n_chan, cor_shape, alpha, freq, freq0):
```

africanus/calibration/tests/test_utils.py
```diff
@@ -97,7 +110,64 @@ def make_dual_pol_data(sigma_n, n_dir, sigma_f):
...
 
-def test_residual_vis():
+@corr_shape_parametrization
+def test_corrupt_vis(corr_shape, jones_shape):
+    """
+    Tests corrupt vis against predict_vis. They should do
+    the same thing but corrupt_vis adheres to the structure
+    in the africanus.calibration module.
+    """
+    np.random.seed(42)
```
sjperkins: The following is something that I have not done throughout the code base, but I would like to start promoting as a pattern for the future: We should start using numpy random states which parametrise the RNG for the test. So something like:

```python
    rs = np.random.RandomState(42)
    vis = rs.normal((rows, chan, corr)) + rs.normal((rows, chan, corr))*1j
    data_dict = calibration_factory(sigma_n, sigma_f, n_time, ..., rs)
```

africanus/calibration/tests/test_utils.py
```diff
@@ -106,9 +176,15 @@ def test_residual_vis():
     np.random.seed(42)
```
sjperkins:
```suggestion
    rs = np.RandomState(42)
```

africanus/calibration/tests/test_utils.py
```diff
@@ -97,7 +110,64 @@ def make_dual_pol_data(sigma_n, n_dir, sigma_f):
...
+    in the africanus.calibration module.
+    """
+    np.random.seed(42)
+    # simulate noise free data with random DDE's
+    n_dir = 3
+    n_time = 32
+    n_chan = 16
+    n_ant = 7
+    sigma_n = 0.0
+    sigma_f = 0.05
```
sjperkins: How about making  some or all of these parametrizations? It would provide the ability to easily try out multiple vales of sigma, for example (0.10 below)

```python
@pytest.mark.parametrize("n_dir", [3])
@pytest.mark.parametrize("sigma_n", [0.0])
@pytest.mark.parametrize("sigma_f", [0.05, 0.10])
def test_corrupt_vis(n_dir, sigma_n, sigma_f, ...):
   data_dict = calibration_factory(sigma_n, sigma_f, ...)
```

africanus/calibration/tests/test_utils.py
```diff
@@ -127,29 +203,34 @@ def test_residual_vis():
...
 
 
-def test_correct_vis():
+@corr_shape_parametrization
+def test_correct_vis(corr_shape, jones_shape):
     """
     Tests correct_vis by correcting noise free simulation
     with random DIE gains
     """
     np.random.seed(42)
```
sjperkins:
```suggestion
    rs = np.RandomState(42)
```

africanus/calibration/tests/test_utils.py
```diff
@@ -26,65 +21,83 @@ def give_lm(n_dir):
...
     assert not np.isnan(model_data).any()
     # simulate gains (just radnomly scattered around 1 for now)
-    jones = np.ones((n_time, n_ant, n_chan, n_dir, n_cor), dtype=np.complex64)
+    jones = np.ones((n_time, n_ant, n_chan, n_dir) +
+                    jones_shape, dtype=np.complex128)
     if sigma_f:
-        jones += sigma_f*(
-            np.random.randn(n_time, n_ant, n_chan, n_dir, n_cor) +
-            1.0j*np.random.randn(n_time, n_ant, n_chan, n_dir, n_cor))
+        jones += (normal(loc=0.0, scale=sigma_f, size=jones.shape) +
```
sjperkins: I think one could use `np.RandomState.normal` here

africanus/calibration/utils.py
```diff
@@ -107,6 +110,42 @@ def jones_inverse_mul(a1j, blj, a2j, out):
...
+
+    @wraps(correct_vis)
+    def _correct_vis_fn(time_bin_indices, time_bin_counts,
+                        antenna1, antenna2, jones, vis, flag):
+        jones_shape = np.shape(jones)
+        n_tim = jones_shape[0]
+        n_dir = jones_shape[3]
+        if n_dir > 1:
+            raise ValueError("Jones has n_dir > 1.\
+                              Cannot correct for direction dependent gains")
```
sjperkins: Note that you can specify a multiline string as follows:

```python
raise ValueError("Jones has n_dir > 1. Cannot correct "
                 "for direction dependent gains")
```

africanus/calibration/utils.py
```diff
@@ -318,7 +395,6 @@ def _residual_vis_fn(time_bin_indices, time_bin_counts, antenna1,
 flag : $(array_type)
```
sjperkins: I think you need to add the ``model`` parameter here:

```python
def residual_vis(time_bin_indices, time_bin_counts, antenna1,
                 antenna2, jones, vis, flag, model):
```

africanus/calibration/utils.py
```diff
@@ -330,3 +406,41 @@ def _residual_vis_fn(time_bin_indices, time_bin_counts, antenna1,
...
+antenna1 : $(array_type)
+    First antenna indices of shape :code:`(row,)`.
+antenna2 : $(array_type)
+    Second antenna indices of shape :code:`(row,)`
+jones : $(array_type)
+    Gains of shape :code:`(time, ant, chan, dir, corr)`
+    or :code:`(time, ant, chan, dir, corr, corr)`.
+model : $(array_type)
+    Model data values of shape :code:`(row, chan, dir, corr)`
+    or :code:`(row, chan, dir, corr, corr)`.
```
sjperkins: ``vis`` needs to be documented here. I would also put it before model in the argument list to be consistent with ``residual_vis``

```python
def corrupt_vis(time_bin_indices, time_bin_counts, antenna1,
                antenna2, jones, model, vis):
```

africanus/calibration/tests/test_phase_only.py
```diff
@@ -6,96 +6,9 @@
...
-    data_dict["DATA"] = vis
-    data_dict["MODEL_DATA"] = model_data
-    data_dict["WEIGHT_SPECTRUM"] = weights
-    data_dict["TIME"] = time
-    data_dict["ANTENNA1"] = antenna1
-    data_dict["ANTENNA2"] = antenna2
-    data_dict["FLAG"] = flag
-    data_dict['JONES'] = jones
-    return data_dict
+from africanus.calibration.tests.test_utils import make_data
```
sjperkins: This can be removed if you implement ``make_data`` as a fixture below

africanus/calibration/tests/test_phase_only.py
```diff
@@ -107,9 +20,15 @@ def test_phase_only_diag_diag():
     np.random.seed(42)
```
sjperkins: There's more of explanation for this suggestion lower down.

```suggestion
    rs = np.RandomState(42)
```

africanus/calibration/tests/test_utils.py
landmanbester: Alright, I had a go at this but not sure if I did it correctly. Maybe you can have a look?

sjperkins:
@landmanbester Shout when you want me to review again.

landmanbester:
Will do. I am busy setting up some examples and then it should be good to go

landmanbester:
Hmmm @sjperkins maybe you can advise me on how to do what I am trying to do [here](https://github.com/landmanbester/codex-africanus/blob/4feb77d524018fe543e143ceb5c296b4f9342071/africanus/calibration/utils/examples/apply_gains_to_ms.py#L88). I have assumed that the visibilities of each source are stored in a separate column of the measurement set and of course I can't assign to dask arrays so I am trying to use da.stack here instead. This seems to work if I do something along the lines of 

import dask.array as da
a1 = da.random.random((1000, 32, 2, 2), chunks=(100, 32, 2,2))
a2 = da.random.random((1000, 32, 2, 2), chunks=(100, 32, 2,2))
a = [a1, a2]
a = da.stack(a, axis=2)

but the resulting array needs to be rechunked along the dir axis since da.stack will set the chunks along the stacked axis to 1 by default. So that is what I am trying to do there but weirdly, when I try this in the example, I get the following error

Traceback (most recent call last):
  File "./apply_gains_to_ms.py", line 88, in <module>
    model = da.stack(model, axis=2)
  File "/home/landman/virtualenvs/codex-africanus/local/lib/python2.7/site-packages/dask/array/core.py", line 3554, in stack
    seq[idx[0]].shape))
ValueError: Stacked arrays must have the same shape. The first 0 had shape (252720, 32, 4), while array 1 has shape (252720, 32, 4)

Any ideas?

sjperkins:
Check that the chunks of the stacked arrays are the same, as well as the dtype?

FYI you can rechunk on a specific axis. Something like this should work:

```python
model = da.stack(model, axis=2).rechunk({2: n_dir})
```

sjperkins:
The following works for me:

```python
ds =  xds_from_ms("/home/sperkins/data/WSRT.MS", group_cols=[])[0]
da.stack([ds.DATA.data, ds.CORRECTED_DATA.data, ds.MODEL_DATA.data], axis=2).rechunk({2: 3}).compute()
```

landmanbester:
> Check that the chunks of the stacked arrays are the same, as well as the dtype?
> 
> FYI you can rechunk on a specific axis. Something like this should work:
> 
> ```python
> model = da.stack(model, axis=2).rechunk({2: n_dir})
> ```

This is very useful voodoo, thanks

landmanbester:
Ok, thanks. It seems to work if I create the xds with
 
xds = xds_from_ms(args.ms, columns=cols, chunks={"row": row_chunks})[0]

it was falling over inside the for loop for some reason. @sjperkins is it possible to create a column before writing to it with xds_to_table? Or do I have to do this with pyrap?

sjperkins:
> @sjperkins is it possible to write to a non-existent column with xds_to_table? Or do I have to create it first?

Yes, in the current version, you'll have to create it first. I'll start working on first class supporting for creating new tables, MS and columns soon.

africanus/calibration/tests/test_utils.py
landmanbester: I will do this if we need more thorough testing of one of the parameters. I am happy with the tests as they are for now

landmanbester:
Actually there is something I don't quite understand viz. why [this](https://github.com/landmanbester/codex-africanus/blob/ed1242776253905a32e9b5c6db8ac353796251b9/africanus/calibration/utils/dask.py#L67) line is necessary. If I don't include it I get an error saying KeyError: 'corr2' and I really don't understand why because I have explicitly defined it above. Also, how does it work when this axis is missing from the inputs?

landmanbester:
Ah damn. I pushed some changes for Charles and broke flake8 compliance. Anyway, I don't want to merge this until I've figured out where the KeyError: 'corr2' comes from. Will try to investigate if I get time this week but @sjperkins don't worry about the review just yet, it's not that urgent.

sjperkins:
> Ah damn. I pushed some changes for Charles and broke flake8 compliance. Anyway, I don't want to merge this until I've figured out where the KeyError: 'corr2' comes from. Will try to investigate if I get time this week but @sjperkins don't worry about the review just yet, it's not that urgent.

Sorry I was so far down in xarray-ms land that I let this slip.

I think you need the `new_axes={"corr2": 2}` in the DIAG case where `corr2` is present in the output but not in the input?

landmanbester:
That was the idea but form some reason I also need it in the FULL case and I'm not sure why. Even weirder is that it works in the DIAG_DIAG case when n_corr=2 but not when n_corr=1. Will try to write better tests to get to the bottom of this but this week I am working mainly on DR imaging with Ming so will have to postpone to when I get back next week

africanus/calibration/phase_only/dask.py
```diff
@@ -0,0 +1,78 @@
...
+    return blockwise(np_compute_jhj, jones_shape,
+                     time_bin_indices, ('row',),
+                     time_bin_counts, ('row',),
+                     antenna1, ('row',),
+                     antenna2, ('row',),
+                     jones, jones_shape,
+                     model, model_shape,
+                     flag, vis_shape,
+                     adjust_chunks={"row": antenna1.chunks[0]},
+                     new_axes={"corr2": 2},  # why?
```
sjperkins: If I recall correctly, this happens because of the dask bug. I'd put a link to the associated dask issue here, or describe it further detail in the comments.

africanus/calibration/phase_only/dask.py
```diff
@@ -0,0 +1,78 @@
...
+                     time_bin_indices, ('row',),
+                     time_bin_counts, ('row',),
+                     antenna1, ('row',),
+                     antenna2, ('row',),
+                     jones, jones_shape,
+                     residual, vis_shape,
+                     model, model_shape,
+                     flag, vis_shape,
+                     adjust_chunks={"row": antenna1.chunks[0]},
+                     new_axes={"corr2": 2},  # why?
```
sjperkins: Document the dask bug

africanus/calibration/phase_only/phase_only.py
```diff
@@ -0,0 +1,395 @@
...
+            out[...] = sign * a1j * blj * a2j.conjugate()
+            # for c in range(out.shape[-1]):
+            #     out[c] = sign * a1j[c] * blj[c] * a2j[c].conjugate()
+    elif mode == DIAG:
+        def jacobian(a1j, blj, a2j, sign, out):
+            out[...] = 0
+    elif mode == FULL:
+        def jacobian(a1j, blj, a2j, sign, out):
+            out[...] = 0
+    return njit(nogil=True)(jacobian)
```
sjperkins: This may now be a candidate for the following:

```suggestion
    return njit(nogil=True, inline='always')(jacobian)
```

africanus/calibration/phase_only/tests/test_phase_only.py
```diff
@@ -0,0 +1,194 @@
...
+    from africanus.calibration.phase_only.phase_only import compute_jhj_and_jhr
+    jhj1, jhr1 = compute_jhj_and_jhr(time_bin_indices, time_bin_counts,
+                                     ant1, ant2, jones, vis, model, flag)
+    jhj2 = compute_jhj(time_bin_indices, time_bin_counts,
+                       ant1, ant2, jones, model, flag)
+    jhr2 = compute_jhr(time_bin_indices, time_bin_counts,
+                       ant1, ant2, jones, vis, model, flag)
+
+    print(np.abs(jhj1-jhj2).max())
+    print(np.abs(jhr1-jhr2).max())
```
sjperkins: The print statements should be removed

africanus/calibration/utils/compute_and_corrupt_vis.py
```diff
@@ -0,0 +1,153 @@
...
+                out[1, 0] += t1*tmp[0, 0] +\
+                    t2*tmp[0, 0] +\
+                    t3*tmp[1, 0] +\
+                    t4*tmp[1, 0]
+                out[1, 1] += t1*tmp[0, 1] +\
+                    t2*tmp[0, 1] +\
+                    t3*tmp[1, 1] +\
+                    t4*tmp[1, 1]
+
+    return njit(nogil=True)(jones_mul)
```
sjperkins: This may be a candidate for inlining

```suggestion
    return njit(nogil=True, inline='always')(jones_mul)
```

africanus/calibration/utils/correct_vis.py
```diff
@@ -0,0 +1,164 @@
...
+            t4 = a11*blj[1, 1]
+            out[1, 0] = t1*b00 +\
+                t2*b00 +\
+                t3*b10 +\
+                t4*b10
+            out[1, 1] = t1*b01 +\
+                t2*b01 +\
+                t3*b11 +\
+                t4*b11
+    return njit(nogil=True)(jones_inverse_mul)
```
sjperkins:
```suggestion
    return njit(nogil=True, inline='always')(jones_inverse_mul)
```

africanus/calibration/utils/corrupt_vis.py
```diff
@@ -0,0 +1,131 @@
...
+FULL = 2
+
+
+def jones_mul_factory(mode):
+    if mode == DIAG_DIAG:
+        def jones_mul(a1j, model, a2j, out):
+            n_dir = np.shape(model)[0]
+            for s in range(n_dir):
+                # for c in range(out.shape[-1]):
+                #     out[c] += a1j[s, c]*model[s, c]*np.conj(a2j[s, c])
```
sjperkins: Remove commented out code.

africanus/calibration/utils/compute_and_corrupt_vis.py
```diff
@@ -0,0 +1,153 @@
...
+
+import numpy as np
+from africanus.util.docs import DocstringTemplate
+from africanus.util.numba import generated_jit, njit
+from .utils import check_type
+from africanus.constants import minus_two_pi_over_c as m2pioc
+
+DIAG_DIAG = 0
+DIAG = 1
+FULL = 2
```
sjperkins: These constants seem to be defined in multiple locations. It would be better if they were placed in a single python module and imported into other locations in the code. Perhaps `africanus.calibration.strategies`?. Or `africanus.constants`

africanus/calibration/utils/correct_vis.py
```diff
@@ -0,0 +1,164 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+from africanus.util.docs import DocstringTemplate
+from africanus.util.numba import generated_jit, njit
+from .utils import check_type
+
+DIAG_DIAG = 0
+DIAG = 1
+FULL = 2
```
sjperkins: Defined in multiple locations

africanus/calibration/utils/correct_vis.py
```diff
@@ -0,0 +1,164 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+from africanus.util.docs import DocstringTemplate
+from africanus.util.numba import generated_jit, njit
+from .utils import check_type
```
sjperkins: Use absolute rather than relative import

```suggestion
from africanus.calibration.utils import check_type
```

africanus/calibration/utils/corrupt_vis.py
```diff
@@ -0,0 +1,131 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+from africanus.util.docs import DocstringTemplate
+from africanus.util.numba import generated_jit, njit
+from .utils import check_type
```
sjperkins: Use absolute rather than relative import

africanus/calibration/utils/corrupt_vis.py
```diff
@@ -0,0 +1,131 @@
...
+
+DIAG_DIAG = 0
+DIAG = 1
+FULL = 2
```
sjperkins: Multiple definitions in multiple files

africanus/calibration/utils/dask.py
```diff
@@ -0,0 +1,234 @@
+# -*- coding: utf-8 -*-
+
+from africanus.calibration.utils.correct_vis import CORRECT_VIS_DOCS
+from africanus.calibration.utils.corrupt_vis import CORRUPT_VIS_DOCS
+from africanus.calibration.utils.residual_vis import RESIDUAL_VIS_DOCS
+from .compute_and_corrupt_vis import COMPUTE_AND_CORRUPT_VIS_DOCS
```
sjperkins: Use absolute import

africanus/calibration/utils/dask.py
```diff
@@ -0,0 +1,234 @@
...
+from africanus.calibration.utils import correct_vis as np_correct_vis
+from . import compute_and_corrupt_vis as np_compute_and_corrupt_vis
```
sjperkins: Use an absolute import

africanus/calibration/utils/dask.py
```diff
@@ -0,0 +1,234 @@
...
+
+    return blockwise(_corrupt_vis_wrapper, out_shape,
+                     time_bin_indices, ("row",),
+                     time_bin_counts, ("row",),
+                     antenna1, ("row",),
+                     antenna2, ("row",),
+                     jones, jones_shape,
+                     model, model_shape,
+                     adjust_chunks={"row": antenna1.chunks[0]},
+                     new_axes={"corr2": 2},  # why?
```
sjperkins: Document the dask bug

africanus/calibration/utils/dask.py
```diff
@@ -0,0 +1,234 @@
...
+    return blockwise(_correct_vis_wrapper, out_shape,
+                     time_bin_indices, ("row",),
+                     time_bin_counts, ("row",),
+                     antenna1, ("row",),
+                     antenna2, ("row",),
+                     jones, jones_shape,
+                     vis, out_shape,
+                     flag, out_shape,
+                     adjust_chunks={"row": antenna1.chunks[0]},
+                     new_axes={"corr2": 2},  # why?
```
sjperkins: Document the dask bug

africanus/calibration/utils/dask.py
```diff
@@ -0,0 +1,234 @@
...
+                     time_bin_indices, ("row",),
+                     time_bin_counts, ("row",),
+                     antenna1, ("row",),
+                     antenna2, ("row",),
+                     jones, jones_shape,
+                     vis, out_shape,
+                     flag, out_shape,
+                     model, model_shape,
+                     adjust_chunks={"row": antenna1.chunks[0]},
+                     new_axes={"corr2": 2},  # why?
```
sjperkins: document the dask bug

africanus/calibration/utils/residual_vis.py
```diff
@@ -0,0 +1,145 @@
...
+
+import numpy as np
+from functools import wraps
+from africanus.util.docs import DocstringTemplate
+from africanus.util.numba import generated_jit, njit
+from africanus.calibration.utils import check_type
+
+DIAG_DIAG = 0
+DIAG = 1
+FULL = 2
```
sjperkins: Defined multiple times in multiple files

africanus/calibration/utils/residual_vis.py
```diff
@@ -0,0 +1,145 @@
...
+                t4 = a1j[s, 1, 1]*model[s, 1, 1]
+                out[1, 0] -= t1*tmp[0, 0] +\
+                    t2*tmp[0, 0] +\
+                    t3*tmp[1, 0] +\
+                    t4*tmp[1, 0]
+                out[1, 1] -= t1*tmp[0, 1] +\
+                    t2*tmp[0, 1] +\
+                    t3*tmp[1, 1] +\
+                    t4*tmp[1, 1]
+    return njit(nogil=True)(subtract_model)
```
sjperkins:
```suggestion
    return njit(nogil=True, inline='always')(subtract_model)
```

africanus/calibration/phase_only/dask.py
```diff
@@ -0,0 +1,78 @@
+
+from .phase_only import COMPUTE_JHJ_DOCS
+from .phase_only import COMPUTE_JHR_DOCS
```
sjperkins: Use absolute imports

africanus/calibration/phase_only/dask.py
```diff
@@ -0,0 +1,78 @@
...
+from africanus.calibration.utils import check_type
+from .phase_only import compute_jhj as np_compute_jhj
+from .phase_only import compute_jhr as np_compute_jhr
```
sjperkins: Use absolute imports

africanus/calibration/utils/__init__.py
```diff
@@ -0,0 +1,7 @@
+# flake8: noqa
+
+from .utils import check_type, chunkify_rows
```
sjperkins: Use absolute imports

africanus/calibration/utils/__init__.py
```diff
@@ -0,0 +1,7 @@
...
+from africanus.calibration.utils.corrupt_vis import corrupt_vis
+from africanus.calibration.utils.correct_vis import correct_vis
+from africanus.calibration.utils.residual_vis import residual_vis
+from .compute_and_corrupt_vis import compute_and_corrupt_vis
```
sjperkins: Use absolute imports

africanus/calibration/utils/compute_and_corrupt_vis.py
```diff
@@ -0,0 +1,153 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+from africanus.util.docs import DocstringTemplate
+from africanus.util.numba import generated_jit, njit
+from .utils import check_type
```
sjperkins: Use absolute imports

africanus/calibration/utils/utils.py
```diff
@@ -0,0 +1,114 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+from africanus.util.docs import DocstringTemplate
+
+DIAG_DIAG = 0
+DIAG = 1
+FULL = 2
```
sjperkins: Define multiple times in multiple files

africanus/calibration/tests/conftest.py
```diff
@@ -0,0 +1,108 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
```
sjperkins: These imports can be removed now that we're python3 only.

africanus/calibration/utils/residual_vis.py
```diff
@@ -0,0 +1,145 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
```
sjperkins: Remove __future__ imports now we're python3

africanus/calibration/utils/tests/conftest.py
```diff
@@ -0,0 +1,108 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
```
sjperkins: Remove future imports now we're python3

africanus/gps/examples/generate_phase_only_gains.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Simulates phase-only gains where phases are drawn from a
+Gaussian process with covariance function given by cov_func
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
```
sjperkins: Remove future imports now we're python3

landmanbester:
@sjperkins do you mind having a look at this before we merge?

africanus/calibration/phase_only/phase_only.py
landmanbester: Will add. I actually did that and removed it while I was trying to figure out why the one test was failing

africanus/calibration/phase_only/dask.py
landmanbester: Thanks, will do

africanus/calibration/utils/compute_and_corrupt_vis.py
landmanbester: I'll put them in utils and import from there since that is where the function lives that checks for them

africanus/calibration/utils/correct_vis.py
landmanbester: Ah man. This was such a neat solution for my awkwardly broken lines when they get too long. Any chance of relaxing the strict 79 character limit?

sjperkins:
Use parentheses to break the line, opening bracket on top line and closing
bracket and imported function on the bottom line

PR #134, opened 2019-08-21
upgrade to dask-ms

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```


sjperkins:
@landmanbester Could you check whether your predict_from_fits example works with the changes I've made? I don't have a handy FITS file to predict from lying around.

sjperkins:
I have however, tested that the scripts start up ;-P

PR #140, opened 2019-08-28
Make predict_vis docstrings consistent

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,7 +5,7 @@ History
 0.1.9 (YYYY-MM-DD)
 ------------------
 
-* Clarify die2_jones/dde2_jones in predict_vis documentation (:pr:`135`)
+* predict_vis documentation improvements (:pr:`135`, :pr:`140`)
 * Upgrade to dask-ms in the examples (:pr:`134`, :pr:`138`)
 * Explain how to obtain predict_vis time_index argument (:pr:`130`)
 * Update RIME predict example to support Tigger LSM's and Gaussians (:pr:`129`)
--- a/africanus/rime/predict.py
+++ b/africanus/rime/predict.py
@@ -508,7 +508,7 @@ to the following formula:
 
 
     V_{pq} = G_{p} \left(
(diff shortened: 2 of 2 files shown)


sjperkins:
@JSKenyon Could you please take a look a the changes and see if they make sense? One issue is that the "base coherencies" is still called **base_vis** but I don't want to rename it throughout the code base.

JSKenyon:
I actually think that the it is the base part of base visibilities that is confusing. If I understood correctly, the matrix B is almost an implementation detail - it expresses the fact that we may accumulate single visibility over a chunked source dimension. In the limit where there is no need to chunk over the source dimension I think that the B matrix becomes unnecessary/zero. Perhaps precomputed visibilities would be a more sensible name. That would also suggest the use case in which B contains model visibilities obtained from another source e.g. a MODEL_DATA column.

sjperkins:
> I actually think that the it is the base part of base visibilities that is confusing. If I understood correctly, the matrix B is almost an implementation detail - it expresses the fact that we may accumulate single visibility over a chunked source dimension. 

Indeed, that is its purpose, but I think the term "coherencies" expresses that which is accumulated more accurately as the DIE terms are not yet applied and would presumably create the wrong result if they were.

> In the limit where there is no need to chunk over the source dimension I think that the B matrix becomes unnecessary/zero. 

Yes

> Perhaps precomputed visibilities would be a more sensible name. That would also suggest the use case in which B contains model visibilities obtained from another source e.g. a MODEL_DATA column.

Is this not a different case as the DIE's have not yet been applied? Something iike:

```python
ds.MODEL_DATA + predict_vis(...)
```

JSKenyon:
As we discussed at lunch, I think base is the misleading term - feel free to settle on an alternative.

PR #145, opened 2019-09-09
Support the beam in the predict example

sjperkins:
- [ x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```


sjperkins:
This PR plugs the codex DDE and Feed Rotation into the predict.py example.

The script confirming agreement with MeqTrees has also been updated.

sjperkins:
@JoshVStaden So the `predict.py` script in master branch now accepts a `--beam` option which will turn on support for holographic beam DDE's.

You should be able to call it as follows:

```bash
$ python predict.py WSRT.MS -b "beam_\$(corr)_\$(reim).fits" -sm sky_model.txt
```

Note the escape of the $ sign. The above pattern refers to eight beam files permuted by two sets `['xx', 'xy', 'yx', 'yy']` and `['re', 'im]`.

I've confirmed that the predict agrees with MeqTrees. So this serves as a reference for the visibilities that should be produced by the Zernike DDEs. I think they'll be slightly off because there's no interpolation along the frequency axis in the Zernike case.

JoshVStaden:
Hi Simon. I have tried running this new script, with a few fits files that I generated from my Zernike script. I understand that beam_cube_dde works by interpolating the beam cube along the frequency axis, but it seems that it falls down when I specify a single frequency (due to a divide by 0 exception). Is this meant to be the case, and the fits file is just assumed to have more than 1 frequency, due to the fact that the frequency is being interpolated?

sjperkins:
Yeah it doesn't do single slices. In fact I would have expected it to fail more sensibly

PR #153, opened 2019-09-19
Spi map example

landmanbester:
Let's get this into master so I can get some feedback. It seems to be working on simulated data at least


africanus/model/spi/examples/simple_spi_fitter.py
```diff
@@ -0,0 +1,402 @@
...
+                   "a - alpha map \n"
+                   "i - I0 map \n"
+                   "I - reconstructed cube form alpha and I0 \n"
+                   "b - interpolated beam \n"
+                   "c - restoring beam used for convolution \n"
+                   "Default is to write all of them")
+    return p
+
+
+args = create_parser().parse_args()
```
sjperkins: I think it would be safer if the contents of the script were placed in a function lilke so:

```python
def main():
   ....

if __name__ == "__main__":
    main()
```

Briefly, this allows the module to be imported without running the script. https://stackoverflow.com/a/19578357/1611416, which can be useful for tools which inspect python code (such as pytest, probably)

africanus/model/spi/examples/README.rst
```diff
@@ -0,0 +1,33 @@
+Simple SPI Fitter
+=================
+
+Fits a simple spectral index model to image cubes. Usage is as follows
+
+$ ./simple_spi_fitter.py --fitsmodel=/path/to/model.fits
```
sjperkins: This is a  nice description of the example. You could introduce some restructured text code formatting here. Either:

```
Fits a simple spectral index model to image cubes. Usage is as follows::

    $ ./simple_spi_fitter.py --fitsmodel=/path/to/model.fits 

```

or 

```
Fits a simple spectral index model to image cubes. Usage is as follows

.. code-block:: bash

    $ ./simple_spi_fitter.py --fitsmodel=/path/to/model.fits 

```
should do it

landmanbester:
Ok, I didn't like the interface to the pyfftw objects and there was a significant overhead to using the pyfftw.interfaces functions so I switched to using pypocketfft. I don't think pypocketfft is on PYPI yet though, you need to install it manually from the repo. @sjperkins do you think we should make this an optional requirement? We will also need it for the gridder and I would like to use it for the multi-scale algorithm I am putting in.
I've also implemented your changes and marked them as resolved but for some reason the tests are not getting triggered on the latest changes. Not sure why. Any ideas?

sjperkins:
> Ok, I didn't like the interface to the pyfftw objects and there was a significant overhead to using the pyfftw.interfaces functions so I switched to using pypocketfft. I don't think pypocketfft is on PYPI yet though, you need to install it manually from the repo. @sjperkins do you think we should make this an optional requirement? We will also need it for the gridder and I would like to use it for the multi-scale algorithm I am putting in.

Are there any codex API's that need pocketFFTW, at present or just the example? If it's just the example, then I don't think it needs to be an optional requirement in the setup.py yet. Rather add it in a gridder/multi-scale PR for e.g.

> I've also implemented your changes and marked them as resolved but for some reason the tests are not getting triggered on the latest changes. Not sure why. Any ideas?

Ah I think this is because I disabled [build on PR](https://docs.travis-ci.com/user/web-ui/#build-pushed-pull-requests) no travis because I was trying to avoid [double builds](https://docs.travis-ci.com/user/pull-requests/#double-builds-on-pull-requests). I've re-enabled this on travis now. I think you may need to push another commit up to trigger this though?

sjperkins:
> Are there any codex API's that need pocketFFTW, at present or just the example? If it's just the example, then I don't think it needs to be an optional requirement in the setup.py yet. Rather add it in a gridder/multi-scale PR for e.g.

Apologies I should have read what you said more closely. We can't put it in setup.py because its not on pypi. I think you should do something like:

```python
try:
   import pocketfft
except ImportError as e:
    opt_import_error = ImportError("Unable to import pocketfft.
                       "Please manually install it from http://pocket.fftw.url/")
else:
    opt_import_error = None


@requires_optional("pocketfft", opt_import_error)
def function(*args, **kwargs):
    ....
```

landmanbester:
@sjperkins looks like we broke test_requirements. Can you have a look please?

sjperkins:
The `len(import_errors) == 0` case in ``_missing_packages`` needs to be catered for, otherwise it returns None.

landmanbester:
Ah yes, of course. I think it should pass now but note that the function now returns a list of packages instead of a tuple. Will that be problem?

africanus/util/requirements.py
```diff
@@ -15,6 +15,9 @@ def _missing_packages(fn, packages, import_errors):
         return ("%s requires installation of "
                 "the following packages: %s.\n"
                 "%s" % (fn, packages, import_err_str))
+    else:
+        return ("%s requires installation of the following packages: %s. "
+                % (fn, *packages))
```
sjperkins: > Ah yes, of course. I think it should pass now but note that the function now returns a list of packages instead of a tuple. Will that be problem?

Do you mean passes a list instead of a tuple?.

I think this should work:

```python
     return ("%s requires installation of the following packages: %s. "
                % (fn, tuple(packages)))
```

landmanbester: Yep. I just changed it to unpack the list and modified the string it compares to accordingly.

PR #154, opened 2019-09-21
Adds bounding hull geometric and image manipulation routines and test

bennahugo:
- [ X] Tests added / passed

As discussed here are the geometric routines. At the moment I don't have time to rePEP8 this but it works with python 3, are reasonably optimized and there is a test included


sjperkins:
@bennahugo I'd like to merge this PR. Do you have any further updates or are you happy with this as it stands?

PR #158, opened 2019-10-02
Modify predict to work with eidos beams

landmanbester:
MOdified predict so it does not insist that beam fits headers have the same DATE entries and to allow the CTYPE's to include px and py since this is how eidos spits them out

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,9 +2,12 @@
 History
 =======
 
+0.2.1 (YYYY-MM-DD)
+------------------
+* Modified predict to be compatible with eidos fits headers (:pr:`158`)
+
 0.2.0 (2019-09-30)
 ------------------
-* Modiefied predict to be compatible with eidos fits headers (:pr:`158`)
 * Added standalone SPI fitter (:pr:`153`)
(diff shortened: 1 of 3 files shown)


HISTORY.rst
```diff
@@ -4,6 +4,7 @@ History
 
 0.2.0 (2019-09-30)
```
sjperkins: Rather put this in a new temporary date placeholder

```rst
0.2.1 (YYYY-MM-DD)
* Modified predict
```

Also correct the Modiefied predict

africanus/util/beams.py
```diff
@@ -158,19 +158,19 @@ def beam_grids(header):
 
     # Find the relevant axes
     for i in range(beam_axes.ndims):
-        if beam_axes.ctype[i] in ('L', 'X'):
+        if beam_axes.ctype[i] in ('L', 'X', 'px'):
```
sjperkins: How about handling both lower case and upper case at once

```suggestion
        if beam_axes.ctype[i].upper() in ('L', 'X', 'PX'):
```

sjperkins:
LGTM, lets merge after the tests succeed

PR #143, opened 2019-08-29
Optimise summation in coherency jones multiply sum

sjperkins:
- Previously used array slicing.
- Explicit indexing is now used.

- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  If the pep8 tests fail, the quickest way to correct
  this is to run `autopep8` and then `flake8` and
  `pycodestyle` to fix the remaining issues.

  ```
  $ pip install -U autopep8 flake8 pycodestyle
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

- [x] Fully documented, including `HISTORY.rst` for all changes
      and one of the `docs/*-api.rst` files for new API

  To build the docs locally:

  ```
  pip install -r requirements.readthedocs.txt
  cd docs
  READTHEDOCS=True make html
  ```

PR #159, opened 2019-10-23
Numba 0.46.0 upgrade

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/africanus/averaging/time_and_channel_avg.py
+++ b/africanus/averaging/time_and_channel_avg.py
@@ -57,29 +57,6 @@ def comp_add_factory(present):
     return njit(nogil=True, cache=True, inline='always')(impl)
 
 
-def sigma_add_factory(have_sigma, have_weight):
-    """
-    Returns function for adding sigma values to a bin.
-    Uses provided weights, else natural weights
-    """
-    if not have_sigma:
-        def impl(out_sigma, out_weight_sum, orow, in_sigma, in_weight, irow):
-            pass
-    elif have_weight:
-        def impl(out_sigma, out_weight_sum, orow, in_sigma, in_weight, irow):
-            for c in range(out_sigma.shape[1]):
-                out_sigma[orow, c] += (in_sigma[irow, c]**2 *
-                                       in_weight[irow, c]**2)
-                out_weight_sum[orow, c] += in_weight[irow, c]
-    else:
-        def impl(out_sigma, out_weight_sum, orow, in_sigma, in_weight, irow):
-            for c in range(out_sigma.shape[1]):
-                out_sigma[orow, c] += in_sigma[irow, c]**2
-                out_weight_sum[orow, c] += in_weight[irow, c]
-
-    return njit(nogil=True, cache=True, inline='always')(impl)
-
-
 def normaliser_factory(present):
     """ Returns function for normalising data in a bin """
     if present:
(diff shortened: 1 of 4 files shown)


sjperkins:
@landmanbester Would you mind taking a look?

numba 0.46.0 now has the ability to [inline](https://numba.pydata.org/numba-doc/latest/developer/inlining.html?highlight=inline) functions. I've utilised this in `predict_vis` as well as the averaging code.

Another thing that's now available in numba is `branch pruning` on compile-time constants. In the example below, the appropriate branch of the if statement will be pruned out at compile-time depending on whether None or an actual value is passed for b.

```python
@njit
def fn(a, b=None):
   if b is None:
       x = 200
   else:
       x = 100
 
   return a * x
```

sjperkins:
The strategies here may also be useful in the calibration code, for example.

africanus/averaging/time_and_channel_avg.py
```diff
@@ -166,25 +49,71 @@ def impl(meta, ant1, ant2, flag_row=None,
...
+                    for co in range(weight.shape[1]):
+                        weight_avg[out_row, co] += weight[in_row, co]
+
+                if sigma is not None:
+                    for co in range(sigma.shape[1]):
+                        sva = sigma[in_row, co]**2
+
+                        # Use provided weights
+                        if weight is not None:
+                            wt = weight[in_row, co]
```
landmanbester: Why are you squaring the weights in the numerator?

sjperkins:
The following test case:

```python
import numpy as np
from africanus.rime.predict import predict_vis


def rf(*a, **kw):
    return np.random.random(*a, **kw)


def rc(*a, **kw):
    return rf(*a, **kw) + 1j*rf(*a, **kw)


def test_benchmark_predict_vis(benchmark):
    s = 10
    t = 100
    a = 14
    c = 64
    co = (2, 2)

    ant1, ant2 = (ar.astype(np.int32) for ar in np.triu_indices(a, 1))
    bl = ant1.shape[0]
    r = t * bl

    a1j = a2j = rc((s, t, a, c) + co)
    blj = rc((s, r, c) + co)
    g1j = g2j = rc((t, a, c) + co)
    bvis = rc((r, c) + co)

    ti = np.repeat(np.arange(t), bl)
    ant1 = np.tile(ant1, t)
    ant2 = np.tile(ant2, t)
    assert ti.shape == ant1.shape

    benchmark.pedantic(predict_vis, warmup_rounds=2, rounds=3,
                       iterations=5,
                       args=(ti, ant1, ant2, a1j, blj, a2j, g1j, bvis, g2j))
```

produces the following timings without inlining:

```
---------------------------------------------------- benchmark: 1 tests ---------------------------------------------------
Name (time in ms)                   Min       Max      Mean  StdDev    Median     IQR  Outliers     OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------
test_benchmark_predict_vis     463.1247  466.6951  464.7775  1.7999  464.5127  2.6778       1;0  2.1516       3           5
---------------------------------------------------------------------------------------------------------------------------
```

and the following times with inling:

```
---------------------------------------------------- benchmark: 1 tests ---------------------------------------------------
Name (time in ms)                   Min       Max      Mean  StdDev    Median     IQR  Outliers     OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------
test_benchmark_predict_vis     135.5459  137.9587  136.5809  0.6175  136.6068  0.4606       2;2  7.3217      10           5
---------------------------------------------------------------------------------------------------------------------------
```
 
so we're looking at ~3.43 times speedup with inlining. In this case, each of the inlined functions does a fair amount of work (2x2 jones matrix multiply) so the speedup is not as dramatic as that described in https://github.com/ska-sa/codex-africanus/pull/143#issuecomment-526231476, which merely involved inlining a summation in an inner loop.

PR #173, opened 2020-01-30
Time and channel BDA

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```


sjperkins:
/cc @o-smirnov  @atemkeng @bennahugo @KshitijT @smasoka  @JSKenyon 

I'm cautiously optimistic that time and channel BDA is occurring in this branch.

Note that currently it is producing sparse flattened data (rowchan, corr), that will be reworked into a (row, chan, corr) MS format.

In the **test_bda_avg** test case, we have a Westerbork antenna layout, combined with a MeerKAT L-band. Typing this out, I suspect a Westerbork band should probably be substituted.

This seems to result in fairly extreme averaging. By this I mean going from 2639 rows and 4096 (659MB) channels  to 376 sparse data points (24KB flattened rows and channels), or a factor of 28748X compression.

Some individual timings below:

```bash
# Time to construct the BDA mapping
mapping: 0.124518s
# Time to average row data
row_average: 0.073361s
# Time to average row and channel data (vis, flag, weight_spectrum, sigma_spectrum)
row_chan_average: 3.288135s
(2639, 4096, 4) 659.75 (376, 4) 24064
```

The above compression factor seems extreme. On the other hand, I believe that Westerbork antenna are closely spaced which may mean that most of the data can be subject to extreme averaging. 8s dumps seem to be averaged together into 240s dumps (30x), while the 4K MeerKAT L-band seems to be averaged down to between 2 and 8 channels (1024X).

While imaging a BDA'd MS would provide the only concrete answer, at this point I'd probably err towards the side of saying that something needs tweaking in the mapping code.

Questions:

1. Can BDA really be that good on this data?
2. Should the test data be made more representative with a WSRT SPW?

sjperkins:
By changing minimum 0.99 decorrelation factor to a 0.999 decorrelation factor in the above example, I get 659MB -> 17MB, or 38X compression factor.

JSKenyon:
Those numbers do seem extreme to me, but that is merely instinct - I have no evidence to back it up. Westerbork does have redundant spacings, but I would still expect decorrelation/smearing over 240 seconds. The same is true for frequency. There are many factors at play here:

- where is the array pointing? Presumably, BDA is affected by the distribution of  UVW samples.
- what is the channel width? I can imagine a fairly dramatic reduction along the frequency axis but I would have thought smearing would place an upper bound on this.
- what is the effect of the array layout? As mentioned, Westerbork has redundant spacings and has more antennas in its "core". Those short spacings will likely be heavily averaged.

As ever, I am not an expert on these things, so take this all with a pinch of salt. As you say, you would need to image the MS to "see" the effect of such extreme averaging. It might be possible to do this in a hacky way by heavily flagging the original MS and then populating what remains with the BDA data.

sjperkins:
Apologies, it's not Westerbork, its some other array of 35 antennae, from which I took the first 14 antennae, confusing myself and possibly others.

sjperkins:
It's actually the first 14 MeerKAT antennae.

atemkeng:
Great, it seems correct to me. For MeerKAT the compression factor is quite big raison is just as you mentioned, most of the baselines are at the core which require more averaging. Combined with wsrt this might even lift up the compression factor as all the wsrt baselines are very short. In a normal regime wsrt would not suffer from decorrelation even if the source is at the edges of the field. As you said, it will be good to see how the flux of the source is spread in the image.

sjperkins:
To be even more precise, it's these antennae:

```
0: m000
1: m003
2: m005
3: m008
4: m009
5: m010
6: m012
7: m014
8: m015
9: m023
10: m025
11: m031
12: m032
13: m033
```

sjperkins:
> Great, it seems correct to me. For MeerKAT the compression factor is quite big raison is just as you mentioned, most of the baselines are at the core which require more averaging. Combined with wsrt this might even lift up the compression factor as all the wsrt baselines are very short. In a normal regime wsrt would not suffer from decorrelation even if the source is at the edges of the field. As you said, it will be good to see how the flux of the source is spread in the image.

Thanks for confirming @atemkeng!

sjperkins:
@smosaka You may want to take a look at the test case outline in https://github.com/ska-sa/codex-africanus/pull/173/commits/374a89e790d991e32a538ba63ea485515ca961b0

sjperkins:
@smasoka @atemkeng @o-smirnov , https://github.com/ska-sa/codex-africanus/pull/173/commits/c254536dc41ea0cf658b4c039151476fbf24b980 contains the changes I was discussing in the meeting.

bennahugo:
For record
```
L := T(RA, DEC)^T * UVW
```
where T is defined as
https://github.com/ska-sa/codex-africanus/blob/master/africanus/gridding/perleypolyhedron/policies/baseline_transform_policies.py#L17

and the synthesis imaging derivatives apply

sjperkins:
As discussed with @bennahugo and @JSKenyon  current implementation using DDFacet + Atemkeng's approximations. Probable bug here in that dt is used to calculate delta psi in the current branch. Current action would be remove to this and retest, first be reverting to the commit here: https://github.com/ska-sa/xova/issues/18#issuecomment-690739544

Longer term:

1. probably worth revisiting the analytic definitions first tried out here: https://github.com/ska-sa/codex-africanus/pull/173/commits/0c4641dac9882289a1e6985b71ee407059efb725, but which are missing several matrix rotations, implemented for the PPGridder [here](https://github.com/ska-sa/codex-africanus/blob/master/africanus/gridding/perleypolyhedron/policies/baseline_transform_policies.py). Transform is L = T(RA,D)^T (uvw).




Use this definition moran and swinson

theta_b = max(lambda) / longest baseline distance 
r = maximum distance in lm
w_e = 15 degrees and hour
delta t = number of seconds worth of data in the bin
erf = scipy.special.erf

for simple check of decorrelation lost i.e. (1 - decorr)

bennahugo:
As discussed the following derivation of the decorrelation factor gives the maximum amount of time a particular baseline can be averaged in a particular direction (RA, DEC) @sjperkins @JSKenyon
image

One can, along similar lines, work out the partial derivative in the radial direction to see what effect averaging lambda has on radial smearing. I would use the norm of the two factors to determine a final smearing level if you want a single knob

sjperkins:
OK, the multiple vis column averaging is now in this branch. Next I'll work on the things mentioned in https://github.com/ska-sa/codex-africanus/pull/173#issuecomment-704832201

PR #189, opened 2020-05-14
Fix incorrect SPI calculation and make predict defaults MeqTree equivalent.

JSKenyon:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/africanus/model/spectral/spec_model.py
+++ b/africanus/model/spectral/spec_model.py
@@ -170,8 +170,8 @@ def spectral_model(stokes, spi, ref_freq, frequency, base=0):
                         spec_model = estokes[s, p]
 
                         for si in range(0, nspi):
-                            term = espi[s, si, p] * freq_ratio**(si + 1)
-                            spec_model += term
+                            term = (freq_ratio + 1) ** espi[s, si, p]
+                            spec_model *= term
 
                         spectral_model[s, f, p] = spec_model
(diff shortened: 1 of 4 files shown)


africanus/model/spectral/spec_model.py
```diff
@@ -170,8 +171,8 @@ def impl(stokes, spi, ref_freq, frequency, base=0):
                         spec_model = estokes[s, p]
 
                         for si in range(0, nspi):
-                            term = espi[s, si, p] * freq_ratio**(si + 1)
-                            spec_model += term
+                            term = (freq_ratio + 1) ** espi[s, si, p]
```
sjperkins, line 174: How about changing the definition of ``freq_ratio`` above to

```python
freq_ratio = frequency[f] / rf
```

This save some FLOPS here:

```suggestion
                            term = freq_ratio**espi[s, si, p]
```

JSKenyon: Done.

JSKenyon:
@sjperkins I believe I have addressed your suggestions.

PR #200, opened 2020-05-28
Support multiple receptor angles in feed_rotation

### LINKED ISSUE #191: Support RECEPTOR_ANGLE in feed_rotation
In the general the P Jones matrix (though I believe it is L Jones in the Codex docs) should also be able to account for the RECEPTOR_ANGLE present in the FEED sub-table. In the most basic of cases, RECEPTOR_ANGLE will be `(n_ant, n_receptor)` where `n_receptor == 2`; one angle for X/R and one angle for Y/L. Doing this at present requires a fair amount of manual intervention but it is absolutely necessary for MeerKAT. I will note that this is only possible if the two numbers specified in RECEPTOR_ANGLE are the same (in general they probably should be), as otherwise we are not dealing with a product of rotation matrices (tagging @bennahugo to confirm).

I would suggest that the [`feed_rotation`](https://github.com/ska-sa/codex-africanus/blob/97b6accde38d0f2d44ffa8743e9be8340c0d5501/africanus/rime/dask.py#L61) function accept an additional parameter, e.g. `feed_offsets`, which are added in the appropriate place. This should preserve the current interface, as the default can be zero, but should make constructing P far less fiddly. Something similar is [done in CubiCal](https://github.com/ratt-ru/CubiCal/blob/e49aff3975d8cc91a29a707c2d60036ad883bc8e/cubical/machines/parallactic_machine.py#L117).

sjperkins:
Closes #191 

- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```


sjperkins:
This is implemented in #239

PR #203, opened 2020-06-11
WIP: update and numbafy linalg module

landmanbester:
I needed to modify and add a few tools for the gain re-interpolation stuff I am adding to cubical. Thus far I have numbafied the required kronecker_tools and added a pre-conditioned conjugate gradient function for solving Hermitian linear systems. I still need to add docs etc. but wanted to start testing


sjperkins:
Cool, note that it's fairly easy to make a point release these days.

PR #209, opened 2020-07-07
Fix Gaussian Shape docs

### LINKED ISSUE #207: Inconsistent documentation for Gaussian shape function
It has just come to my attention that the formulae in the docs (https://codex-africanus.readthedocs.io/en/latest/model-api.html) do not match up with what is actually computed in the function 

https://github.com/ska-sa/codex-africanus/blob/d4e4c01de1029e418f03c5fb096829a3202986b2/africanus/model/shape/gaussian_shape.py#L49

Look at the expression for v_1 compared to how it is defined in the code. Also, the docs refer to lambda as frequency and I am pretty sure it should be wavelength. There is also no mention of the scaling function that compensates for the fact that e_{min} and e_{maj} actually refer to the FWHM of the components (which is not the usual way to parametrise a Gaussian). It might be wise to update the docs to explain this parameterization a bit better.

sjperkins:
Fixes #207 

- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/africanus/model/shape/gaussian_shape.py
+++ b/africanus/model/shape/gaussian_shape.py
@@ -60,18 +60,18 @@ Computes the Gaussian Shape Function.
 
 .. math::
 
-    & \lambda^\prime = 2 \lambda \pi \\
+    & \nu^\prime = 2 \nu \pi \\
     & r = \frac{e_{min}}{e_{maj}} \\
     & u_{1} = (u \, e_{maj} \, cos(\alpha) - v \, e_{maj} \, sin(\alpha))
-      r \lambda^\prime \\
+      r \nu^\prime \\
     & v_{1} = (u \, e_{maj} \, sin(\alpha) - v \, e_{maj} \, cos(\alpha))
-      \lambda^\prime \\
-    & \textrm{shape} = e^{(-u_{1}^2 - v_{1}^2)}
+      \nu^\prime \\
+    & \textrm{shape} = e^{(-(u_{1}^2) - (v_{1}^2))}
 
 where:
 
 - :math:`u` and :math:`v` are the UV coordinates and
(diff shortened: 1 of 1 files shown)


sjperkins:
@landmanbester Based on #207, I've updated the docs to look as follows:

Screenshot from 2020-07-07 13-29-25.

I haven't described the convention for the Gaussian Shape Parameters. Can you point me at a reference for their specification?

sjperkins:
Ignore my previous comment, I need to put the scaling factor in.

sjperkins:
Documentation now looks like:

Screenshot from 2020-07-07 14-00-27

Do you have a reference for the emaj/emin/position angle conventions?

landmanbester:
Sorry, I think I pointed at the wrong line. In the code, v1 is defined as

https://github.com/ska-sa/codex-africanus/blob/d4e4c01de1029e418f03c5fb096829a3202986b2/africanus/model/shape/gaussian_shape.py#L45

So there is a plus instead of the minus as indicated in the docs. Also, the effect of multiplying u1 by the ratio r is simply to replace emaj with emin so I am not sure what the point of r really is. Is this parametrisation correct? I would have to work out what the general parameterization of a 2D Gaussian is in Fourier space, don't know it off the top of my head. It seems to be correct for

. 

Also, we should document the fact that emaj and emin refer to the FWHM of the Gaussian not the standard deviation to clarify the factor of

landmanbester (approved):
This looks correct, I think. Kind of lost the context

PR #210, opened 2020-07-08
Test binner class

sjperkins:
- [ ] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```


sjperkins:
@smasoka I've rebased your **test_Binner_class** branch ontop of the **time-and-channel-bda** branch.

This will make it easier to integrate your test cases into #173, which in turn can be merged into master when ready.

There are some conflicts, but they should be easy enough to resolve by merging **time-and-channel-bda** into **test_Binner_class**

sjperkins:
binner test case extensions

- Select a small subset of existing test data fixtures OR setup a small test data subset that exercises the code paths described below

- Then test the following functions in the following ways:

- Binner.start_bin
- Binner.add_row cases:
   - auto correlation case
   - adding row causes decorrelation case (returns False)
   - adding row does not decorrelation case (returns True)
- Binner.finalise_bin cases:
   - bin contains single value case.
   - auto corrrelation case.
   - bin contains multiple values.

PR #219, opened 2020-08-26
Use Sentry Literals in the Perley Polyhedron Gridder

sjperkins:
- [ ] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  If the pep8 tests fail, the quickest way to correct
  this is to run `autopep8` and then `flake8` and
  `pycodestyle` to fix the remaining issues.

  ```
  $ pip install -U autopep8 flake8 pycodestyle
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

- [ ] Fully documented, including `HISTORY.rst` for all changes
      and one of the `docs/*-api.rst` files for new API

  To build the docs locally:

  ```
  pip install -r requirements.readthedocs.txt
  cd docs
  READTHEDOCS=True make html
  ```

PR #222, opened 2020-09-23
Reintroduce Perley-Polyhedron Gridder

bennahugo:
Reverts ska-sa/codex-africanus#220


bennahugo:
Need to try and undo the mess created by the revert. I will be basing off this branch to make the necessary workaround for the numba 0.5 bug

o-smirnov:
Nothing like a revert of a revert to pervert one's brain!

sjperkins:
@bennahugo Would you not prefer to work from #219 whose point was to retain the state of the PR prior to the revert?

sjperkins:
Whatever works though, this has green lights on the test suite.

bennahugo:
@sjperkins that has rather large changes with SanitaryLiterals. I think the issue is not in the use of SanitaryLiterals but in the way the new numba compiler handles literals. It seems you cannot create a stack of nested literals in a jit function called by another jit function. All ```literally``` calls must take place in the first jit function the compiler encounters on the stack, so I strongly suspect this is a compiler bug in numba >=0.5

sjperkins:
> @sjperkins that has rather large changes with SanitaryLiterals. I think the issue is not in the use of SanitaryLiterals but in the way the new numba compiler handles literals. It seems you cannot create a stack of nested literals in a jit function called by another jit function. All `literally` calls must take place in the first jit function the compiler encounters on the stack, so I strongly suspect this is a compiler bug in numba >=0.5

All good!

sjperkins:
I think the numba version needs changing in the setup.py?

bennahugo:
woops sorry you are right. I only upgraded locally

sjperkins:
Well that was easy enough! Green lights. Let me give it a quick once over, unless there's more you wish to do?

bennahugo:
Sure go ahead. The issue is simplified to the simplest possible reproducer that mimics this issue in https://github.com/numba/numba/issues/6274

sjperkins:
@bennahugo I'm a bit confused. Were you intending to post this on the related issue #219?

This is the PR reintroducing the gridder.

sjperkins:
Happy to review though.

bennahugo:
> @bennahugo I'm a bit confused. Were you intending to post this on the related issue #219?
> 
> This is the PR reintroducing the gridder.

No, I just mentioned the post where I've reproduced the workaround for compiler bug that causes the failure on numba version 0.51.2 in a simpler case if you wanted to comment

sjperkins:
> > @bennahugo I'm a bit confused. Were you intending to post this on the related issue #219?
> > This is the PR reintroducing the gridder.
> 
> No, I just mentioned the post where I've reproduced the workaround for compiler bug that causes the failure on numba version 0.51.2 in a simpler case if you wanted to comment

Ah! I'll take a look at that too.

HISTORY.rst
```diff
@@ -6,6 +6,8 @@ History
 ------------------
 * Add wrapper for ducc0.wgridder (:pr:204`)
 * Correct Irregular Grid nesting in BeamAxes (:pr:`217`)
+* Add Perley Polyhedron Faceting Gridder/Degridder (:pr:`202`, :pr:`215`)
```
sjperkins, line 9: Could you start a new version section in the changelog that also mentions this PR? e.g.:

```
0.2.7 (YYYY-MM-DD)
----------------------------
* Add Perley Polyhedron Faceting Gridder/Degridder (:pr:`202`, :pr:`215`, :pr:`222`)

```

africanus/gridding/perleypolyhedron/kernels.py
```diff
@@ -0,0 +1,178 @@
...
+
+@requires_optional('scipy', scipy_import_error)
+def kbsinc(W, b=None, oversample=5, order=15):
+    """
+    Modified keiser bessel windowed sinc (Jackson et al.,
+    IEEE transactions on medical imaging, 1991)
+    with a modification of higher order bessels as default, as
+    this improves the kernel at low number of taps.
+    """
+    autocoeffs = np.polyfit(
```
sjperkins, line 55: Could the definition of this be moved outside the kbsinc function?

africanus/gridding/perleypolyhedron/kernels.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def hanningsinc(W, a=0.5, oversample=5):
+    """
+    Basic hanning windowed sinc
+    """
+    autocoeffs = np.polyfit([1.5, 2.0, 2.5, 3.0, 3.5],
+                            [0.7600, 0.7146, 0.6185, 0.5534, 0.5185], 3)
+    if a is None:
+        a = np.poly1d(autocoeffs)((W + 2))
```
sjperkins, line 75: Similarly to kbsinc, could the autocoeffs be moved outside?

africanus/gridding/perleypolyhedron/kernels.py
```diff
@@ -0,0 +1,178 @@
...
+        b = np.poly1d(autocoeffs)((W + 2))
+
+    u = uspace(W, oversample)
+    wnd = jn(order, b * np.sqrt(1 - (2 * u /
+                                     ((W + 2) + 1))**2)) * 1 / ((W + 2) + 1)
+    res = sinc(W, oversample=oversample) * wnd * np.sum(wnd)
+    return res / np.sum(res)
+
+
+def hanningsinc(W, a=0.5, oversample=5):
```
sjperkins, line 68: Shouldn't None be the default value for ``a``?

```suggestion
def hanningsinc(W, a=None, oversample=5):
```

africanus/gridding/perleypolyhedron/tests/test_daskintrf.py
```diff
@@ -0,0 +1,367 @@
...
+                               c(d0)],
+                              [c(d0) * c(h0), -c(d0) * s(h0),
+                               s(d0)]])
+                uvw[n * ntime + ih0, :] = np.dot(R, blpos[n, :].T)
+        uvw = da.from_array(uvw, chunks=(row_chunks, 3))
+        pxacrossbeam = 5
+        nchan = 128
+        frequency = da.from_array(np.linspace(1.0e9, 1.4e9, nchan),
+                                  chunks=(nchan, ))
+        wavelength = 299792458.0 / frequency
```
sjperkins, line 75: It's possible to use the lightspeed definition from constants

```
from africanus.constants import c as lightspeed
```

africanus/gridding/perleypolyhedron/tests/test_daskintrf.py
```diff
@@ -0,0 +1,367 @@
...
+                               oversample=OS)
+    nrow = int(5e4)
+    uvw = np.column_stack(
+        (5000.0 * np.cos(np.linspace(0, 2 * np.pi, nrow)),
+            5000.0 * np.sin(np.linspace(0, 2 * np.pi, nrow)), np.zeros(nrow)))
+
+    pxacrossbeam = 10
+    nchan = 1024
+    frequency = np.array(np.linspace(1.0e9, 1.4e9, nchan))
+    wavelength = np.array([299792458.0 / f for f in frequency])
```
sjperkins, line 218: The ``np.array`` and list comprehensions aren't strictly necessary:

```python
from africanus.constants import c as lightspeed

frequency = np.linspace(1.0e9, 1.4e9, nchan)
wavelength = lightspeed / frequency
```

africanus/gridding/perleypolyhedron/tests/test_daskintrf.py
```diff
@@ -0,0 +1,367 @@
...
+    nrow = int(5e4)
+    nrow_chunk = nrow // 32
+    uvw = np.column_stack(
+        (5000.0 * np.cos(np.linspace(0, 2 * np.pi, nrow)),
+            5000.0 * np.sin(np.linspace(0, 2 * np.pi, nrow)), np.zeros(nrow)))
+
+    pxacrossbeam = 10
+    nchan = 1024
+    frequency = np.array(np.linspace(1.0e9, 1.4e9, nchan))
+    wavelength = np.array([299792458.0 / f for f in frequency])
```
sjperkins, line 270: The ``np.array`` and list comprehensions aren't strictly necessary:

```python
from africanus.constants import c as lightspeed

frequency = np.linspace(1.0e9, 1.4e9, nchan)
wavelength = lightspeed / frequency
```

africanus/gridding/perleypolyhedron/tests/test_daskintrf.py
```diff
@@ -0,0 +1,367 @@
...
+    nrow = 100
+    nrow_chunk = nrow // 8
+    uvw = np.column_stack(
+        (5000.0 * np.cos(np.linspace(0, 2 * np.pi, nrow)),
+            5000.0 * np.sin(np.linspace(0, 2 * np.pi, nrow)), np.zeros(nrow)))
+
+    pxacrossbeam = 10
+    nchan = 16
+    frequency = np.array(np.linspace(1.0e9, 1.4e9, nchan))
+    wavelength = np.array([299792458.0 / f for f in frequency])
```
sjperkins, line 324: The ``np.array`` and list comprehensions aren't strictly necessary:

```python
from africanus.constants import c as lightspeed

frequency = np.linspace(1.0e9, 1.4e9, nchan)
wavelength = lightspeed / frequency
```

sjperkins (changes requested):
Some minor nits, but otherwise this is good to go.

sjperkins:
@bennahugo One final request: Could you rename the **lambdas** arguments to **wavelengths** in the gridder and degridder calls? I think its more explicit.

africanus/gridding/perleypolyhedron/kernels.py
bennahugo: Don't think so it is specific for the function (the coefficents are not the same - see the original Jackson paper).

africanus/gridding/perleypolyhedron/kernels.py
bennahugo: well spotted yes will fix

africanus/gridding/perleypolyhedron/kernels.py
sjperkins: Ok, but you could have _kb_coeffs and _hanning_coeffs globals, for instance. They're constant and private by convention.

bennahugo:
Fixes the issues raised apart from the coefficients which are function specific so I don't think they should be moved outside the function?? uvlambda is a common expression for stating uv measured in lambda - which is what was meant in the gridder interface. I've changed it though

sjperkins (approved):

sjperkins:
Squashed into master. I'm very glad this is now in!

I'm aiming to push a release out today.

PR #230, opened 2020-11-06
Specify epsilon for the wgridder manually

landmanbester:
There was a bug in wgridder.dirty which caused the gridder to crash if an epsilon other than the default was used. This should now be fixed. I have also changed the default to 1e-5 regardless of whether single or double precision griding is required.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,10 @@
 History
 =======
 
+0.2.9 (????-??-??)
+------------------
+* Fix manually specifying wgridder precision (:pr:`230`) 
+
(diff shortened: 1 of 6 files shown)


sjperkins (changes requested):
Looks good, just needs a HISTORY.rst entry.

landmanbester:
Yes, I still want to add a test or two as well

landmanbester:
I think it's good to go in now if you have a sec @sjperkins

sjperkins (approved):

sjperkins:
Thanks @landmanbester

PR #234, opened 2021-01-04
Add mindet to avoid div0 errors in spi fitter and fix alpha and I0 variance estimates

landmanbester:
Small fix to avoid divide by zero errors during spi fitting. Let's hold off merging for now, want to test this thoroughly first

### DIFF
--- a/africanus/model/spi/component_spi.py
+++ b/africanus/model/spi/component_spi.py
@@ -10,7 +10,7 @@ from africanus.util.numba import jit
 
 @jit(nopython=True, nogil=True, cache=True)
 def _fit_spi_components_impl(data, weights, freqs, freq0, out,
-                             jac, ncomps, nfreqs, tol, maxiter):
+                             jac, ncomps, nfreqs, tol, maxiter, mindet):
     w = freqs/freq0
     for comp in range(ncomps):
         eps = 1.0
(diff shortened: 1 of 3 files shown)


landmanbester:
@sjperkins I think this is good to go when you have a sec

landmanbester:
Oh wait, need the HISTORY.rst entry

landmanbester:
Just to document the changes, this PR fixes the variance estimates of alpha and I0 when the per band weights in the fit are only correct up to some scaling factor and adds a minimum cutoff to the determinant computation that can cause the GN optimisation to fail if starting too far away from a local minimum

africanus/model/spi/component_spi.py
```diff
@@ -31,23 +32,23 @@ def _fit_spi_components_impl(data, weights, freqs, freq0, out,
             jr0 = 0.0
             jr1 = 0.0
             for v in range(nfreqs):
-                lik += residual[v] * weights[v] * residual[v]
+                lik += residual[v] * weights[v] * residual[v]  # /2
```
sjperkins, line 35: Can this comment be removed?

sjperkins (approved):
I'm happy, I'll leave it up to you to decide whether you want to deal with comment nit.

africanus/model/spi/component_spi.py
landmanbester: It has been removed

PR #237, opened 2021-01-25
Add hessian function and double accumulation option to gridder

landmanbester:
* adds a paranoid mode to the gridder to avoid accumulation errors when gridding in single precision. 
* adds the Hessian function (out of memory and in place) R.H W R to wgridder.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,7 @@ History
 
 0.2.10 (YYYY-MM-DD)
 -------------------
+* Add double accumulation option and Hessian function to wgridder (:pr:`237`)
 * Upgrade ducc0 to version 0.8.0 (:pr:`236`)
 * Add mindet to avoid div0 errors in spi fitter and fix alpha and I0 variance
   estimates (:pr:`234`)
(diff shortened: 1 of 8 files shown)


landmanbester:
Alright, ready for you when you have a sec @sjperkins. As discussed I haven't included any faceting capabilities in this PR

sjperkins (commented):
Looks good to me. Do you think its worth adding test cases for this functionality?

landmanbester:
Yes, definitely. I was just trying to come up with a good test for the functionality that won't add too much to the runtime. I think I might have just thought of one. Let's hold off for a bit before merging

landmanbester:
Alright, I have added a test for double accumulation while simultaneously testing that hessian produces the dirty image when acting on the model. There is only a single parametrisation for the test because it takes a while (need a large number of rows to test double accumulation). Note that the hessian function allows us to compute residual images as

IR = ID - hessian(x)

where, since the hessian doesn't require visibilities, we have effectively reduced the IO by two thirds (only need to read in (assumed fixed) real valued weights).

sjperkins:
In! thanks @landmanbester  and @mreineck

PR #238, opened 2021-01-26
Support beam model during SPI fitting

landmanbester:
* Adds per component beam to the model during spi fitting
* ensure only a single ref_freq_index is returned in component_spi function as discussed here https://github.com/cyriltasse/DDFacet/pull/729


landmanbester:
Should be good to go

sjperkins (approved):

PR #240, opened 2021-02-05
Set dof in spi fitter to min of 1

landmanbester:
As reported here

https://github.com/ratt-ru/pfb-clean/issues/40

we need to make sure that the effective dof during spi fitting does not go to zero.


landmanbester:
That was fat. Thanks @sjperkins

landmanbester:
*fast

sjperkins:
Quick fix!

PR #246, opened 2021-04-12
[skip ci] Interpolation missing nominal values during Averaging

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/africanus/averaging/bda_mapping.py
+++ b/africanus/averaging/bda_mapping.py
@@ -168,8 +168,6 @@ class Binner(object):
         self.rs = row
         self.re = row
         self.bin_count = 1
-        self.time_sum = time[row]
-        self.interval_sum = interval[row]
         self.bin_flag_count = (1 if flag_row is not None and flag_row[row] != 0
                                else 0)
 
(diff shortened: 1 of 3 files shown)


sjperkins:
@bennahugo would you consider giving this a quick review, mostly of the new test cases: `test_interpolation` and `test_interpolation_edge`? I'm less interested in review of the refactor, although you're welcome to provide feedback here too. Regardless I think the new test cases illustrate how the code now interpolates missing rows, as well as how it deals (imperfectly) with missing edge data.

Note that this doesn't change the BDA averaging code, yet. It only modifies the standard time and channel averaging code because it's far easier to reason about including X seconds worth of data in the bin. If you're happy with the standard case, I'll implement the same in the BDA code.

sjperkins:
Discussion Points
------
* Regular/Irregular grid
* Missing/Non-missing data


10 seconds bin
3 seconds
X is missing (not even flagged)
O is present (flagged/unflagged)

O O O X X X X O  

| O O O | X X X X O  
| O O O | X X X | X O |     -- split samples across multiple bins

| O O O | X X X X O |       -- current strategy, samples completely in/out of a bin

| O O O | X X X X | O |

PR #269, opened 2022-01-28
LazyProxy and LazyProxyMultiton patterns

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,11 @@
 History
 =======
 
+X.Y.Z (YYYY-MM-DD)
+------------------
+* Add Multiton, LazyProxy and LazyProxyMultiton patterns (:pr:`269`)
+
+
 0.3.2 (2022-13-01)
 ------------------
 * Support numba >= 0.54 (:pr:`264`)
--- a/africanus/experimental/rime/fused/specification.py
+++ b/africanus/experimental/rime/fused/specification.py
@@ -1,7 +1,7 @@
 import ast
 from importlib import import_module
 import inspect
-from itertools import groupby  # noqa
+import multiprocessing
 from pathlib import Path
 import re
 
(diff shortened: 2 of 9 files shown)


sjperkins:
@JSKenyon @bennahugo @o-smirnov 

The purpose of this PR is to demonstrate two useful patterns for managing resources in a dask context. I'll briefly describe the `LazyProxy` which is constructed with the function/class and arguments for constructing an instance of the class. 

The instance is only created when attempts to access attributes on it via the `LazyProxy` occur. Additionally, `LazyProxies` are lightweight to pickle, assuming that the supplied arguments are lightweight. This means that they can be embedded in dask graphs without incurring creation of expensive resources like files or database connections and transferred to dask workers.

```python
f = LazyProxy(open, "test.txt", mode="w")
f.write("Hello World")
f.close()
```

Additionally, it's possible to supply a `finaliser` method to the `LazyProxy`, which can be used to finalise the instance when the `LazyProxy` is garbage collected. For example.


```python
def finalise(file_):
  file_.close()

f = LazyProxy((open, finalise), "test.txt", mode="w")
f.write("Hello World")
```

`LazyProxyMultiton` takes this one step further by always returning the same instance if the arguments are the same:

```python
assert LazyProxy(open, "test.txt", mode="w") is not LazyProxy(open, "test.txt", mode="w")
assert LazyProxyMultiton(open, "test.txt", mode="w") is LazyProxyMultiton(open, "test.txt", mode="w")
```

I would like to discuss these patterns over the next couple of weeks in a meeting.

sjperkins:
It works for submitting parangle computation to to "spawn" Process Pools

JSKenyon:
Thanks @sjperkins for taking the time to demonstrate this pattern. I can confirm that this works, with the caveat that in the distributed case it is still necessary to set `dask.config.set({"distributed.worker.daemon": False})`. Dask does seem to detect more leaked semaphores in this case (if the code is interrupted) but seems to clean them up without issue.

PR #276, opened 2023-04-06
Adjust SPI code to handle negative/zero Stokes components.

JSKenyon:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,7 @@ History
 
 0.3.4 (2023-10-03)
 ------------------
+* Adjust SPI code to handle negative/zero Stokes components (:pr:`276`)
 * Separate stokes and correlation dimensions in dask fused RIME (:pr:`273`)
 * Disallow feed rotation terms for RIME's containing less than four correlations (:pr:`273`)
 * Update trove hash (:pr:`274`)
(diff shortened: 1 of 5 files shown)


sjperkins (changes requested):
Thanks for putting this fix in. Could you change the following test case to use the `log` base?

https://github.com/ratt-ru/codex-africanus/blob/7a756a9bd3fe8bfadb2a02add9b3f180be42696d/africanus/experimental/rime/fused/tests/test_rime.py#L250-L253

There are other cases that use the standard base and it would be nice to cover this case too.



https://github.com/ratt-ru/codex-africanus/blob/master/africanus/experimental/rime/fused/tests/test_rime.py

JSKenyon:
I have made the requested changes and propagated the fixes to the other SPI evaluation code.

sjperkins (changes requested):
This looks good. Can you add the following entry to HISTORY.rst? "Adjust SPI code to handle negative/zero Stokes components".

I suggest prefixing the commit with "[skip ci]" to avoid kicking off the CI build. i.e. something like "[skip ci] Update HISTORY.rst".

sjperkins (approved):

PR #281, opened 2023-05-19
Pin numba until such time @overload can be implemented

### LINKED ISSUE #280: Pin release version of numba to < 0.59.0
Codex version: 0.3.4

Raised downstream by crystalball. I see generated_jit is being removed in the near future so I'm marking this and make a patch release of codex to ensure the downstream packages don't start failing -- I don't think there is time now to modify the codebase to use @overload. Per docs we should pin to to below that version.

bennahugo:
Closes #280 

- [ ] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,7 @@ jobs:
         run: pipenv graph
 
       - name: Run complete test suite
-        run: pipenv run py.test --flake8 -s -vvv africanus/
+        run: NUMBA_NRT_STATS=1 pipenv run py.test --flake8 -s -vvv africanus/
 
   deploy:
     needs: [test]
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ if not on_rtd:
         # astropy breaks with numpy 1.15.3
         # https://github.com/astropy/astropy/issues/7943
         "numpy >= 1.14.0, != 1.15.3",
-        "numba >= 0.53.1"
+        "numba >= 0.53.1, <0.59.0"
     ]
 
 extras_require = {


sjperkins:
Thanks for looking at his @bennahugo. Pinning numba seems to be a sensible interim measure.

> Hmm tests are currently not working. Setting NUMBA_NRT_STATS appears not to enable NRT counters. Will need further digging

 One thing that immediately springs to mind is the following fixture https://github.com/ratt-ru/codex-africanus/blob/9f6836a5402673bedda2035956a3786dc1ddc61b/africanus/conftest.py#L9-L17 which was designed to check for NRT allocations in the Fused RIME. If the NRT interface changed in numba, this might be the culprit for the test failures.

I'd leave it for Monday though!

.github/workflows/ci.yml
```diff
@@ -79,7 +79,7 @@ jobs:
         run: pipenv graph
 
       - name: Run complete test suite
-        run: pipenv run py.test --flake8 -s -vvv africanus/
+        run: NUMBA_NRT_STATS=1 pipenv run py.test --flake8 -s -vvv africanus/
```
sjperkins, line 82: I don't think `NUMBA_NRT_STATS=1` gets inherited by the  virtual environment created by pipenv


```suggestion
        run: pipenv run NUMBA_NRT_STATS=1 py.test --flake8 -s -vvv africanus/
```

In a fresh python 3.9 virtual environment the test suite is running:

```bash
$ NUMBA_NRT_STATS=1 py.test --flake8 -s -vvv africanus/
```

sjperkins:
Related to:

- https://github.com/numba/numba/pull/8235

bennahugo:
Superceeded by #284

PR #284, opened 2023-05-24
Pin numba to < 0.59 due to pending @generated_jit deprecation

sjperkins:
Closes #280 

Supercedes:
 - #281 

Related to:
 - #283 

- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  If the pep8 tests fail, the quickest way to correct
  this is to run `autopep8` and then `flake8` and
  `pycodestyle` to fix the remaining issues.

  ```
  $ pip install -U autopep8 flake8 pycodestyle
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

- [x] Fully documented, including `HISTORY.rst` for all changes
      and one of the `docs/*-api.rst` files for new API

  To build the docs locally:

  ```
  pip install -r requirements.readthedocs.txt
  cd docs
  READTHEDOCS=True make html
  ```

PR #286, opened 2023-05-25
Optionally check NRT allocations

sjperkins:
Related to:
 - #284

The global `check_allocations` pytest fixture checks for leaks in internal Numba intrinsic code, but without `NUMBA_NRT_STATS` enabled, all tests will fail.

This change only enables the `check_allocations` fixture if `NUMBA_NRT_STATS` is positively set. This means that a developer can run `py.test` without the entire test suite failing. The downside is that numba intrinsic memory leaks might remain locally undiscovered, but these should be picked up by CI where `NUMBA_NRT_STATS=1 py.test` is called.

- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/africanus/conftest.py
+++ b/africanus/conftest.py
@@ -11,6 +11,7 @@ from africanus.util.testing import mark_in_pytest
 NUMBA_NRT_STATS_ENABLED = (
     os.environ.get("NUMBA_NRT_STATS", "false").lower() in {"true", "1"})
 
+
 @pytest.fixture(scope="function", autouse=NUMBA_NRT_STATS_ENABLED)
 def check_allocations():
     """ Check allocations match frees """
--- a/africanus/conftest.py
+++ b/africanus/conftest.py
@@ -1,12 +1,17 @@
 # -*- coding: utf-8 -*-
 
+import os
+
 from numba.core.runtime import rtsys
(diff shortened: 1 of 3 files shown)


africanus/conftest.py
```diff
@@ -1,12 +1,18 @@
...
+import os
+
 from numba.core.runtime import rtsys
 import pytest
 
 from africanus.util.testing import mark_in_pytest
 
 
-@pytest.fixture(scope="function", autouse=True)
+NUMBA_NRT_STATS_ENABLED = (
```
JSKenyon, line 11: Is this the best way of doing this? It should be possible to do `numba.config.NRT_STATS`, but not sure of the context here.

JSKenyon (approved):
Looks good to me - my only comment was for my own interest more than anything else.

africanus/conftest.py
sjperkins: Yes, thats much nicer, thanks for pointing that out.

PR #287, opened 2023-07-04
Calibration Utility Updates

landmanbester:
- [ ] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  ```
  $ autopep8 -r -i africanus
  $ flake8 africanus
  $ pycodestyle africanus
  ```

  To build the docs locally:

  ```
  cd docs
  READTHEDOCS=True make html
  ```

Add a check to make sure the correlation axis is never larger than 2

### DIFF
--- a/africanus/calibration/utils/utils.py
+++ b/africanus/calibration/utils/utils.py
@@ -39,15 +39,21 @@ def check_type(jones, vis, vis_type='vis'):
     return mode
 
 
-def chunkify_rows(time, utimes_per_chunk):
+def chunkify_rows(time, utimes_per_chunk, daskify_idx=False):
     utimes, time_bin_counts = np.unique(time, return_counts=True)
     n_time = len(utimes)
+    if utimes_per_chunk == 0 or utimes_per_chunk == -1:
+        utimes_per_chunk = n_time
     row_chunks = [np.sum(time_bin_counts[i:i+utimes_per_chunk])
                   for i in range(0, n_time, utimes_per_chunk)]
     time_bin_indices = np.zeros(n_time, dtype=np.int32)
     time_bin_indices[1::] = np.cumsum(time_bin_counts)[0:-1]
     time_bin_indices = time_bin_indices.astype(np.int32)
     time_bin_counts = time_bin_counts.astype(np.int32)
+    if daskify_idx:
+        import dask.array as da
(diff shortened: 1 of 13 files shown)


africanus/calibration/utils/utils.py
```diff
@@ -39,15 +39,23 @@ def check_type(jones, vis, vis_type='vis'):
...
     time_bin_indices = np.zeros(n_time, dtype=np.int32)
     time_bin_indices[1::] = np.cumsum(time_bin_counts)[0:-1]
     time_bin_indices = time_bin_indices.astype(np.int32)
     time_bin_counts = time_bin_counts.astype(np.int32)
+    if daskify_idx:
+        import dask.array as da
+        time_bin_indices = da.from_array(
+            time_bin_indices, chunks=utimes_per_chunk)
+        time_bin_counts = da.from_array(
+            time_bin_counts, chunks=utimes_per_chunk)
```
sjperkins, line 58: Hehe, I'm so on the fence about this. On the one hand it's probably very convenient, on the other hand it feels like this dask wrapping functionality should be elsewhere. A `dask_chunkify_rows` wrapper around `chunkify_rows` might be "better"? But I don't feel strongly about this, its a utility function after all so happy to go with whatever you decide here.

africanus/gps/kernels.py
```diff
@@ -33,13 +33,13 @@ def exponential_squared(x, xp, sigmaf, l, pspec=False):  # noqa: E741
     if pspec:
         N, D = x.shape
         if D != 1:
-            raise NotImplementedError("Only 1D pspecs supported")
+            raise (NotImplementedError, "Only 1D pspecs supported")
```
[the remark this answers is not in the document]
landmanbester: Oops, not sure how that slipped in

africanus/calibration/utils/utils.py
landmanbester: I have removed it, was just getting repetitive but I agree it's better to do it externally

sjperkins:
I forgot to ask for the update to `HISTORY.rst`. If you prefix the commit with `[skip ci]` it'll avoid another ci run.

landmanbester:
Oops, forgot about that. Thanks @sjperkins

PR #288, opened 2023-09-13
remove new_axes={'corr2': 2} in calibration utils

landmanbester:
This was originally required because of a bug in blockwise which has now been fixed (see https://github.com/dask/dask/issues/5550). Using correct_vis in DIAG_DIAG mode failed for me hence the change (not sure why the tests didn't catch that). All seems to work without the new_axes in blockwise


sjperkins (approved):
LGTM, pending the CI successfully running through.

PR #292, opened 2024-01-30
Address warnings

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  virtuale environment:

  ```bash
  $ pip install pre-commit
  $ pre-commit install
  $ pre-commit run -a
  ```

  To build the docs locally:

  ```bash
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,11 +41,17 @@ jobs:
         with:
           fetch-depth: 1
 
+      - name: Create a .env file
+        run: >
+          touch .env
+          echo NUMBA_NRT_STATS=1 >> .env
+          echo NUMBA_CAPTURED_ERRORS='new_style' >> .env
+
       - name: Install base codex-africanus
         run: pipenv install .[testing]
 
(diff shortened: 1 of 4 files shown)


africanus/gridding/perleypolyhedron/policies/stokes_conversion_policies.py
```diff
@@ -5,7 +5,7 @@ def stokes2corr(vis_in, vis_out, policy_type):
     pass
 
 
-@overload(stokes2corr, inline="always")
+@overload(stokes2corr, inline="always", prefer_literal=True)
```
sjperkins, line 8: @bennahugo Just letting you know that the test cases you wrote easily highlighted up the above needed change, due to the following https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-old-style-numba-captured-errors.

So the test cases were useful to have and I didn't have to understand the implementation to do the upgrade.

PR #307, opened 2024-05-23
Fix bda overload to return an implementation

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  virtuale environment:

  ```bash
  $ pip install pre-commit
  $ pre-commit install
  $ pre-commit run -a
  ```

  To build the docs locally:

  ```bash
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/africanus/averaging/bda_avg.py
+++ b/africanus/averaging/bda_avg.py
@@ -3,6 +3,7 @@
 from collections import namedtuple
 
 import numpy as np
+from numba import types
 
 from africanus.averaging.bda_mapping import bda_mapper, RowMapOutput
 from africanus.averaging.shared import chan_corrs, merge_flags, vis_output_arrays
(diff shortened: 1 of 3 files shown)


africanus/averaging/bda_avg.py
```diff
@@ -757,72 +758,132 @@ def nb_bda_impl(
...
+            meta.chan_width,
+            meta.flag_row,
+            row_avg.antenna1,
+            row_avg.antenna2,
+            row_avg.time_centroid,
+            row_avg.exposure,
+            row_avg.uvw,
+            row_avg.weight,
+            row_avg.sigma,
+            # None,  # chan_data.chan_freq,
```
landmanbester, line 876: Is there a reason why chan_freq is not returned?

sjperkins: Good question. Part of the reason is that it technically becomes a (row, chan) dimension as the frequencies and channel widths can also vary per row. This sort of data is available meta object, higher up in this namedtuple construction. There's a chan_width member with (row, chan) dimensionality. However, there isn't a `chan_freq` member. Do you need it? I suppose it wouldn't be necessary if one were only doing time BDA.

sjperkins: Ah there's already an issue

- https://github.com/ratt-ru/codex-africanus/issues/263

landmanbester: I should add that when doing time only BDA you shoudn't need uvw to be expanded as rowchan. This actually uses quite a bit of memory. It's not a train smash for now but it might be worth adding a time only BDA function in the future that doesn't do this

landmanbester (approved):
LGTM. I've tested this on some simulations and at least the time only case seems to work as expected. Thanks @sjperkins with that we now have BDA support in pfb

PR #309, opened 2024-05-27
Detect auto-correlations via np.allclose, rather than equality

sjperkins:
- [x] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  virtuale environment:

  ```bash
  $ pip install pre-commit
  $ pre-commit install
  $ pre-commit run -a
  ```

  To build the docs locally:

  ```bash
  cd docs
  READTHEDOCS=True make html
  ```

### DIFF
--- a/africanus/averaging/bda_mapping.py
+++ b/africanus/averaging/bda_mapping.py
@@ -515,7 +515,7 @@ def nb_bda_mapper(
             binner.reset()
 
             # Auto-correlated baseline
-            auto_corr = ubl[bl, 0] == ubl[bl, 1]
+            auto_corr = np.allclose(ubl[bl, 0], ubl[bl, 1])
 
             for t in range(ntime):
                 # Lookup row, continue if non-existent


landmanbester:
Unfortunately this didn't solve the problem. Dropping flagged rows (the auto-corrs are flagged in this case) seems to get me around the error though

sjperkins:
> Unfortunately this didn't solve the problem. Dropping flagged rows (the auto-corrs are flagged in this case) seems to get me around the error though

Another way this bug could manifest is if `max_fov=0.0`, but by default this is set to `max_fov=3.0`.

landmanbester:
I've left it to the default in this case. Let me try to create a reproducer then we take it from there

landmanbester (commented):
This should not really be necessary. I've been using the master for BDA without issues so I think this PR can be closed

landmanbester:
Apologies, I see I am still using the workaround of removing flagged rows. I will test without it and get back to you

sjperkins:
I feel that an allclose check can't hurt!

PR #336, opened 2025-07-25
Modify convert for purely complex outputs and allow scalar Stokes to multiple correlation mappings

landmanbester:
- [ ] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  virtuale environment:

  ```bash
  $ pip install pre-commit
  $ pre-commit install
  $ pre-commit run -a
  ```

  To build the docs locally:

  ```bash
  cd docs
  READTHEDOCS=True make html
  ```

----
📚 Documentation preview 📚: https://codex-africanus--336.org.readthedocs.build/en/336/

### DIFF
--- a/africanus/model/coherency/cuda/tests/test_convert.py
+++ b/africanus/model/coherency/cuda/tests/test_convert.py
@@ -15,25 +15,25 @@ from africanus.model.coherency.tests.test_convert import (
 
 
 @pytest.mark.skip
-def test_stokes_schemas(in_type, input_schema, out_type, output_schema, vis_shape):
+def test_stokes_schemas(input_schema, output_schema, vis_shape):
     input_shape = np.asarray(input_schema).shape
     output_shape = np.asarray(output_schema).shape
 
-    vis = visibility_factory(vis_shape, input_shape, in_type)
+    vis = visibility_factory(vis_shape, input_shape)
     xformed_vis = np_convert(vis, input_schema, output_schema)
     assert xformed_vis.shape == vis_shape + output_shape
 
 
 @pytest.mark.parametrize(
-    "in_type, input_schema, out_type, output_schema",
+    "input_schema, output_schema",
     stokes_corr_cases + stokes_corr_int_cases,
 )
 @pytest.mark.parametrize("vis_shape", vis_shape)
-def test_cuda_convert(in_type, input_schema, out_type, output_schema, vis_shape):
+def test_cuda_convert(input_schema, output_schema, vis_shape):
     cp = pytest.importorskip("cupy")
 
     input_shape = np.asarray(input_schema).shape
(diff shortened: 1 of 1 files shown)


landmanbester:
This might seem a little bit dirty to you @sjperkins but it seems to be working. Open to suggestions on how to do it better. I need this functionality in the `degrid` worker in `pfb`

africanus/model/coherency/conversion.py
```diff
@@ -11,32 +11,34 @@
...
     },
     "V": {
-        ("XY", "YX"): lambda xy, yx: (xy - yx).imag / 2,
-        ("RR", "LL"): lambda rr, ll: (rr - ll).real / 2,
+        ("XY", "YX"): lambda xy, yx: (xy - yx) / 2j,
+        ("RR", "LL"): lambda rr, ll: (rr - ll) / 2,
     },
 }
 
+corr_products = ["XX", "YX", "XY", "YY", "RR", "RL", "LR", "LL"]
```
sjperkins, line 40:
```suggestion
CORR_PRODUCTS = {"XX", "YX", "XY", "YY", "RR", "RL", "LR", "LL"}
```

africanus/model/coherency/conversion.py
```diff
@@ -11,32 +11,34 @@
 from africanus.util.docs import DocstringTemplate
 
 stokes_conv = {
```
sjperkins, line 13:
```suggestion
STOKES_CONV = {
```

sjperkins (changes requested):
Yeah current implementation is a bit messy (but works) so I think some refinement is necessary

What about making the constants more explicit with something like `["I", "Q=0", "U=0", "V=0"] -> ["XX", "XY", "YX", "YY"]`

landmanbester:
It seems to be working with your changes @sjperkins although I must admit I find the new version far more difficult to read. I'll bow to your wisdom on such matters though. I added a small test for the 'I' to all correlations case and one that ensures we get a failure when requesting 'I' from a single correlation. I think it's ready for another review

africanus/model/coherency/tests/test_convert.py
```diff
@@ -51,7 +51,7 @@ def visibility_factory(vis_shape, input_shape, in_type, backend="numpy", **kwarg
     else:
         raise ValueError("Invalid backend %s" % backend)
 
-    if in_type == "real":
+    if in_type == "complex":
```
sjperkins, line 54: Maybe in_type is redundant, now that the tests aren't distinguishing between complex and real? Could it be removed from the cases and `visibility_factory`?

sjperkins (commented):
Thanks for adding the tests and getting the dask variant to work. I suspect that the `in_type` parameter is now redundant and can be removed.

sjperkins (approved):

PR #338, opened 2025-08-29
Improve input checking and raising of TypingErrors during RIME construction

sjperkins:
- [ ] Tests added / passed

  ```bash
  $ py.test -v -s africanus
  ```

  virtuale environment:

  ```bash
  $ pip install pre-commit
  $ pre-commit install
  $ pre-commit run -a
  ```

  To build the docs locally:

  ```bash
  cd docs
  READTHEDOCS=True make html
  ```

----
📚 Documentation preview 📚: https://codex-africanus--338.org.readthedocs.build/en/338/

### DIFF
--- a/africanus/experimental/rime/fused/core.py
+++ b/africanus/experimental/rime/fused/core.py
@@ -49,8 +49,8 @@ def rime_impl_factory(rime_spec, ncorr):
             raise TypingError(
                 f"Compiling the RIME failed for specification {rime_spec} failed. "
                 f"This often results from passing incorrectly shaped or typed inputs "
-                f"for the various Terms. "
-                f"Please search the resulting stack trace for this error,"
+                f"to the various Terms and Transformers or bugs in their implementations. "
+                f"Please search the associate stack trace for the error"
             )
 
         if not len(args) % 2 == 1:


africanus/experimental/rime/fused/examples/custom_brightness_term.py
```diff
@@ -0,0 +1,113 @@
...
+            "model_flux": ("source", "chan", "stokes"),
+        }
+
+    def init_fields(self, typingctx, init_state, model_flux):  # (ndir, nchan, nstokes)
+        if not isinstance(model_flux, types.Array) or model_flux.ndim != 3:
+            raise TypingError(
+                f"model_flux {model_flux} should be a (ndir, nchan, nstokes) array"
+            )
+
+        return [], lambda init_state, model_flux: None
```
sjperkins, line 41: This simplified a bit (I think in the original version we were defining a function (rather than just a dummy lambda).

Actually if `model_flux` is a requested input, it gets placed on the `state` object in the sampler anyway (I'd forgotten this), so there was no need to do this ourselves

landmanbester (approved):
The example is working for me now, thanks @sjperkins. Looks like mostly type checking, lots of it! I'll approve with the caveat that I don't fully understand all the intrinsics stuff

